PR #652, opened 2014-10-13
Dirty.fix.variable.conflicts.in primdec

jakobkroeker:
fix for three variable conflicts;
the fix for `zeroRad` is dirty but readable.
(compare with the less readable conflict-free one https://github.com/surface-smoothers/Sources/commit/c256fed67c289a6af0b48a9d819394403888c28d)

### DIFF
--- a/Singular/LIB/primdec.lib
+++ b/Singular/LIB/primdec.lib
@@ -747,6 +747,7 @@ static proc splitCharp(list l)
   def P=basering;
   int i,j,k,m,q,d,o;
   int n = nvars(basering);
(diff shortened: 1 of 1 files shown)


hannes14:
fixed in different (better) way via ringlist and without a double ring change

jakobkroeker:
> fixed in different (better) way via ringlist and without a double ring change

Agree, the solution with ringlist may be better if done properly.

but the affected code parts are not fixed yet:
Line 1355 - not fixed
Line 1373 - not fixed
Line 4194 - not fixed
line 7461 - not fixed

PR #653, opened 2014-10-14
fix normalP for zero and unit ideal input, singular trac 596

jakobkroeker:
fixed `normalP()` for zero ideal input.
`normalP()` should now also behave similar to `normal()` for the unit ideal case.
(please review/check!)

open task: update documentation for normal() and normalP() for the unit ideal case 
 will be done in a different pull request.


hannes14:
The groebner computation at the beginning is very expensive (and not necessary).
A better solution is fe88bc9d3dafa5333c8bb5558f5e4513582d55cc.

PR #661, opened 2014-10-24
added an extended test for trac issue 658 (necessary!)

jakobkroeker:


hannes14:
reason for the bug was missing code - it is now there, no need to clutter the test suite.

jakobkroeker:
@malex984 
Hans wrote
> reason for the bug was missing code - it is now there, no need to clutter the test suite.

The IS need for test, since the issue was fixed once, then it gets broken again, 
so without a test this might happen a THIRD time

PR #669, opened 2014-11-11
Walk stuff

oberfran:


hannes14:
under discussion locally, will be solved directly.

PR #673, opened 2014-11-21
replace 'std(..,hilb,..)' calls by 'groebner()' calls in 'primdec.lib'

jakobkroeker:
replace std(..,hilb,..) calls by groebner() calls in 'primdec.lib'

reason:  additional a posteriori check (which?) is required, which is existant in groebner(), 
but is usually missing in 'primdec.lib' code.

Symptoms:
bugs like http://www.singular.uni-kl.de:8002/trac/ticket/662

### DIFF
--- a/Singular/LIB/primdec.lib
+++ b/Singular/LIB/primdec.lib
@@ -2362,6 +2362,7 @@ static proc minAssPrimes_i(int patchPrimaryDecomposition, ideal i, list #)
     list re=imap(P,re);
     return(re);
   }
+
   q = facstd(i);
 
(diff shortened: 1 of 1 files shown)


hannes14:
This would throw away valuable information - it is a bad idea (TM).

jakobkroeker:
>This would throw away valuable information - it is a bad idea (TM).

Can you guarantee that the affected code is sane?

It is also a bad idea too to keep bugs in the source,  
see http://www.singular.uni-kl.de:8002/trac/ticket/662

PR #686, opened 2015-01-08
normaliz.lib support for gradings

csoeger:
The version in singular and normaliz were not synchronized. This is a merged version.
For me unclear points:
* It seems like the Singular team has removed "@texinfo" tags, but then the affected parts are not displayed correctly in html. Maybe you can have a look if it is ok this way.
* Do you have a version number system? It seems like it simply started with 4.0.0.0


hannes14:
The version of a libary should (in the first 3 place) coincide with the version of the Singular kernel
(currently 4.0.1). Before 4.0.0 subversion was used to write the version into the libary which is no longer possible with git, used since 4.0.0.

PR #695, opened 2015-02-08
generation of the manpages with help2man within the autools chain

jgmbenoit:
This sequence of patches allows to generate the {,T,E}Singular manpages with help2man within the autotools chain: the first patch sligtly modiies the their implemented help message to render them readable by help2man; the second patch add an hidden option that literally dump Singular version; the third patch implements the help2man machinery within the autotools chain; the fourth patch wipes out the now obsoleted manpages.


jgmbenoit:
Please note that the current manpages were generated, certainly by hand a long time ago, with help2man (see the commented header).

hannes14:
The problem I am having with this pull request: it fails horribly if help2man does not exist.
It should (in this case) either install a pre-computed man page or do not install a man page at all,
but not fail.
(We try to keep the number of required tools a small as possible)

jgmbenoit:
I see your point. As it concerns only the autotools part, would you be ready to commit a least the two first patches ? Concerning the autotools part, it can be encpasulated withing the maintainer mode.

hannes14:
First 2 patches merged.

PR #697, opened 2015-02-12
chg: allow users to pass dp and ds as orderings + code cleanup suggested by Hans and Oleksandr

YueRen:
bbfan.cc
- moved initialization and deletion temp1 outside of loop to avoid unnecessary initializations / deletions
- commented out function coneContaining, as it is not used (and does not work yet)

tropical.cc
- added function "lowerHomogeneitySpace"

tropicalStrategy.cc
- allow users to use dp or ds or ws in basering when calling "tropicalVariety(ideal I)"
- removed unnecessary function
- removed out of place comments
- removed redundant standard bases computation

tropicalVariety.cc
- removed redundant standard bases computation


YueRen:
P.S. your fixes did work, thanks Oleksandr!

PR #703, opened 2015-03-11
renamed : algemodstd.lib  to nfmodstd.lib

tigistdereje:


hannes14:
- Tst/Long/nfmodstd_l.res.gz must not be contained in the patch
- please change also all other places:Tst/Long/ok_l2.lst, Tst/Short.lst, doc/singular.doc

PR #704, opened 2015-03-17
fix: memleak in nlInitMPZ

YueRen:
setting "z->s = 3" represents that "z->n" is unitialized and hence it is ignored in the destructor.


YueRen:
P.S.: This was not tested on TC, can anybody do that, please?

PR #708, opened 2015-03-20
chg: removed redundant std warning

YueRen:
removed std warning when reducing with respect to a polynomial if the basering is no quotient ring

### DIFF
--- a/Singular/iparith.cc
+++ b/Singular/iparith.cc
@@ -3055,7 +3055,8 @@ static BOOLEAN jjREAD2(leftv res, leftv u, leftv v)
 }
 static BOOLEAN jjREDUCE_P(leftv res, leftv u, leftv v)
 {
-  assumeStdFlag(v);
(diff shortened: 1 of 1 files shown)


hannes14:
No, computing the normal form of a polynomial wrt. an ideal requires always
that the ideal is given as a standard basis. For a qring: I+Q must be a standard basis
(that is the definition of being a standardbasis in a qring).

YueRen:
I see, it seems as if I misunderstood what the jjREDUCE_P does. 

I'll give it another shot then, hopefully right this time. The goal is to suppress these warnings when reducing with respect to a polynomial in a polynomial ring (i.e. no quotient ring).

hannes14:
This is difficult as there is no function that does this (reducing wrt. a polynomial).
Set option(warn) to what is happening: the polynomial will be converted to an ideal
and an result of an automatic type conversion does not have attributes.

PR #711, opened 2015-04-17
chg: min and max now take more than 2 inputs

YueRen:
this has been bothering me for a while, changed min and max to take more than 2 inputs.


malex984:
what about tests and examples?

YueRen:
Added some more examples. Not sure if tests are necessary, since max and min have to tests to begin with.

jakobkroeker:
> Not sure if tests are necessary, since max and min have to tests to begin with.

nonsense.
The interface changed, so please add corresponding tests.

PR #712, opened 2015-04-22
GradedModules.lib from Hanieh Keneshlou<hkeneshlou@yahoo.com>  recieved on 2015.04.22

malex984:
NOTE that i had to correct the recieved input from Hanieh in order to make it usable for Singular.

Resulting HTML manual is available at 
http://www.mathematik.uni-kl.de/~motsak/agr/docs/gradedModules.html

@hannes14: Please don't merge this PR before a positive review by Hanieh.


malex984:
ok, i have just updated and pushed it with ... e5e219020eee81e7dc5562da6a0aa0a4a1e2b3e3

PR #717, opened 2015-07-03
Added rAssure_c_dp(ring r)

adipopescu:


kernel/combinatorics/hdegree.cc
```diff
@@ -1030,6 +1030,10 @@ void scComputeHC(ideal S, ideal Q, int ak, poly &hEdge, ring tailRing)
     hComp(hexist, hNexist, k, hexist, &hNstc);
   else
     hNstc = hNexist;
+  if (hNexist==0)
```
hannes14: Why this? If hNexist would be 0, this routine must not be called anyway:
hEdge would be undefined!
Proposal: change to assume(hNexist>0);

libpolys/polys/kbuckets.cc
```diff
@@ -13,7 +13,7 @@
 //#include <kernel/pShallowCopyDelete.h>
 #include <coeffs/coeffs.h>
 #include <polys/monomials/ring.h>
-//#include <kernel/p_Procs.h>
```
hannes14: This has no effect and libpolys/polys/templates/p_Procs.h must not be used outside of
the template stuff!

libpolys/polys/monomials/p_polys.cc
```diff
@@ -4529,8 +4529,8 @@ static inline unsigned long GetBitFields(const long e,
 // if (ev1 & ~ev2) then exp1 does not divide exp2
 unsigned long p_GetShortExpVector(const poly p, const ring r)
 {
-  assume(p != NULL);
```
hannes14: p_GetShortExpVector is one of the routine that must not be called with p==NULL.
Returning 0 is a reallay bad idea as 0 stand here for p being a NON-ZERO- constant polynomial.
If this routine is called with p==NULLm the caller is wrong!

libpolys/polys/monomials/p_polys.cc
```diff
@@ -4529,8 +4529,8 @@ static inline unsigned long GetBitFields(const long e,
...
   if (p == NULL) return 0;
```
adipopescu: This line was added in this commit. I simply changed the order of assume and this line. 
f3094ae7 (Oleksandr Motsak  2012-07-23 17:29:18 +0200 4532)   if (p == NULL) return 0;

I will leave it as before my commit.

libpolys/polys/kbuckets.cc
adipopescu: Understood

kernel/combinatorics/hdegree.cc
adipopescu: i will change it with your proposal.

adipopescu:
These commits are already in #718

PR #719, opened 2015-07-04
fix for zz is number warning

jakobkroeker:
'`zz` is ..'- warning causes problems for me in testing (others are ok)
It would be nice if this change could be merged in.
Thanks,

Jakob
```
LIB("primdec.lib");
ring R = (0),(x,y,z),lp;
ideal I = -197*y^2*z^2+67;
option(warn);
minAssGTZ(I);
// ** `zz` is poly in primdec.lib::newZero_decomp:9025:  int @k,@s,@n,@k1,zz;


ring S = (19,vv),(x,y,z),lp;
minpoly = (vv^2+6*vv-4);
ideal I = (5*vv-4)*y;
option(warn);
primdecGTZ(I);
// ** `zz` is number in primdec.lib::zero_decomp:876:  int @k,@s,@n,@k1,zz;
```

### DIFF
--- a/Singular/LIB/primdec.lib
+++ b/Singular/LIB/primdec.lib
@@ -873,7 +873,7 @@ EXAMPLE: example zero_decomp; shows an example
   def   @P = basering;
   int uytrewq;
   int nva = nvars(basering);
-  int @k,@s,@n,@k1,zz;
+  int @k,@s,@n,@k1,@zz;
   list primary,lres0,lres1,act,@lh,@wh;
   map phi,psi,phi1,psi1;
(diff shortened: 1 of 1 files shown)


hannes14:
option(warn) is a tool for optimization and debugging - it should not switched on for normal computations (incl. testing). You would also not run your production server via gdb - would you?

jakobkroeker:
> You would also not run your production server via gdb

No, I would not. 
Wha I would is enable assertions for testing and also some warnings from 'option(warn)' .
Since I can't really enable only warnings I'm interested in,
 I enable all of them and disable a small warning subset by a patch.

PR #720, opened 2015-07-13
allow modules in Modstd::reduce_parallel

steenpass:


malex984:
Looks ok, but what about tests for such a usage?

steenpass:
Tests will be provided by Dereje via an implementation of modStd() for modules.

PR #725, opened 2015-09-08
Trac739

mmklee:
fix for trac 739


jakobkroeker:
Thanks! That was amazingly fast!

PR #727, opened 2015-10-05
add new lib: GND.lib

adipopescu:


hannes14:
- SEE ALSO- entry is syntactically invalid: must be a node name of the manual
  (use REFERENCES)
- finde: that is sat(ideal(0),i)[2]
- inY: that is variables(p) contains only var(nrx).....,var(nry)
- myjet: avoid execute, use ringlist
- invp: avoid execute, use ringlist
- invp: if p!=0, then always leadcoef(p)!=0 (line 360): should be: if (ord(p)!=0)
- invp: that is (in principle) division(1,p,bound)
- findminor: avoid execute, use (line 416): resf[size(resf)+1]=intvec(total2[i]1..n]); etc.
- mymin: use min
- findnewvar: see safeVarName (from for example deRham.lib)
- mylcm: avoid execute, use ringlist
- desingularization: option(notWarnSB) is dangerous
- desingularization: void execute, use ringlist

adipopescu:
done the changes. Questions:

* inY: that is variables(p) contains only var(nrx).....,var(nry)

changed the discription. Is there an easier way to do it?

* desingularization: option(notWarnSB) is dangerous

Why? I could get rid of it but before each reduce, i have to set the flag for ideal accordingly

*desingularization: void execute, use ringlist

Done. Just two executes remain there: the ones where a build an ideal from a list. Any workarounds?

hannes14:
- inY: the result of variables(p) should help
- the result of several commands assume that the input is a SB. You get a warning, if Singular does not know that the input is a SB: this may have 2 reasons: 
A: it is not a SB - then you have to change your code
B: it is a SB for some other reason - then at least a comment (why) should be added and
    the warning may be diabled locally (by setting ist isSB attribute)

adipopescu:
Done.
* disabled the option, set used std in cases
* inY is a function that i don't use in my code, but it seems i am already working with variables

jakobkroeker:
@adipopescu 
I recommend - no, I INSIST that appropriate regression tests are designed and implemented before the library is merged in. Doing that 'later' significantly  increases  the risk that it will never happen. (I can easily give several examples in Singular where regression tests are completely missing).

See also 
https://github.com/Singular/Sources/wiki/Step-by-Step-Instructions-for-Contributing-to-Singular

Even more important, I predict that writing tests will result in a positive experience: 
bugs will be catched in an early state, lot of wasted time for debugging will be saved 
(with tests it is often much easier and faster to localize a newly introduced bug)
and the library users will be happier.

The problem is, to get the positive experience  it is necessary to change. And to cite 'Peopleware': 
```
"People hate change…
and that’s because people hate change…
I want to be sure that you get my point.
People really hate change.
They really, really do."
```
!!!

adipopescu:
added tests.

PR #730, opened 2015-11-04
neutralize: tests for Singular/dyn_modules/syzextra

jgmbenoit:
It appears that the test scripts for Singular/dyn_modules/syzextra are not present in the last source ball 4.0.2p2: this patch neutralizes them from the Makefile.am itself.


hannes14:
see f2078ebbc640cc92f43acffa7079852974936c36: we better include these test scripts

Issue #732, opened 2015-11-04
omalloc fails test on arm64 and powerpc architectures


jgmbenoit:
According to the Debian Package Auto-Building machinery, omalloc fails test on arm64 and powerpc architectures [1]: can you reproduce the issue and eventually fix it ?

[1] https://buildd.debian.org/status/package.php?p=singular

jgmbenoit:
I just tried to reproduced on an arm64 Debian porter: I observed not test failure on this porter. Do you have any idea about this issue ?

malex984:
AFAIK `make check` for omalloc may fail due to _unlucky_ random number generation :(

jgmbenoit:
How to generate LUCKY random number  ?

jgmbenoit:
Indeed, failures occur randomly, at least on the arm64 porter.
I have just observed that on the powerpc architecture log file, the error is a segment-fault error: are we also unlucky here ?

malex984:
AFAIK omalloc testing/verification may oversee some memory error in some random cases, which fails `make check`. Segment-fault is not supposed to happen in any case. Which CPU exactly is that? Does segment-fault happen there all the time?

jgmbenoit:
The segment-fault happened on powerpc arch [2]. Whether it happens all the time, right now I cannot say because I have no direct acces to this box. Otherwise, I am on my way to write a getlucky shell simple wrapper that may answer this question and get us lucky.

https://buildd.debian.org/status/fetch.php?pkg=singular&arch=powerpc&ver=4.0.2-p2%2Bds-1&stamp=1446660425

jgmbenoit:
Patch #733 cannot fix the segmentation-fault observed on powerpc architecture [3].
Can you reproduce and eventually fix this segmentation-fault ?
 
[3] https://buildd.debian.org/status/fetch.php?pkg=singular&arch=powerpc&ver=4.0.2-p2%2Bds-2&stamp=1446768703

jgmbenoit:
Finally I get access to a powerpc porter box: there is a firework of Segmentation faults.
Please let me know how I can help here.

malex984:
Could you please run the faulty executable (within /omalloc) under any debugger and send us its output together with backtrace? Do you need step-by-step guidance how to do that?

Would it be possible to build & test the debug version? 
See https://github.com/Singular/Sources/wiki/Building-Singular-from-source#debug-version-of-singular

jgmbenoit:
[...]

PR #733, opened 2015-11-05
tests: omalloc: getlucky

jgmbenoit:
omalloc testing/verification may oversee some memory error in some random cases:
this patch introduces a simple wrapper that is meant to work around those unlucky cases by running a few number of tests instead of just one test.


hannes14:
Solved by changing configure of omalloc. (see issue #732)

jgmbenoit:
This issue #733 is distinct from the issue #732. This issue means that the test is not deterministic, this is not an expected behaviour.

PR #752, opened 2016-03-14
fix: dim int crash

adipopescu:


hannes14:
What about:
ring r1=(integer,2,4),x,dp; dim(std(ideal(x)));
ring r2=(integer,10),(x,y),dp;dim(std(ideal(x)));
(gives sometimes 2, sometimes 1)

PR #753, opened 2016-03-16
Customstd

YueRen:
Fehlt noch die Doku, der Eintrag sollte wie folgt aussehen, aber so soll es hin?

@c --------------------------------------
@node satstd,,
@subsubsection 
satstd                                                    

@cindex satstd

@table @code
@item @strong{Syntax:}
@code{satstd(} ideal I @code{)}
@*@code{satstd(} ideal I, ideal J @code{)}
@item @strong{Type:}
ideal
@item @strong{Assume:}
J is generated by variables
@item @strong{Purpose:}
Starts a standard basis computation in which elements are freely divided by the generators of J.
Hence returns a standard basis of an ideal between I and the saturation I:J^\infty. 
In particular, if I is saturated with respect to J, returns a standard basis of I.

If J is unspecified, it is assumed to be the ideal generated by all variables in the base ring.
@item @strong{Example:}
@smallexample
@c example
LIB "customstd.so";
ring r = 0,(x,y,z),dp;
ideal I = x2+x,y2+y,z2;
satstd(I,ideal(x));
satstd(I,ideal(x,y));
satstd(I,ideal(z));
@c example
@end smallexample
@end table


Singular/dyn_modules/customstd/customstd.cc
```diff
@@ -0,0 +1,482 @@
...
+    }
+    omFree(mm);
+    omFree(m0);
+  }
+  return b; // return TRUE if sp was changed, FALSE if not
+}
+
+// returns 1<=i<=rVar(r) if x is the i-th variable,
+// return 0 otherwise
+static int getVariableIndex(const poly x, const ring r)
```
hannes14: Please consider unsing p_Var instead

Singular/dyn_modules/customstd/customstd.cc
```diff
@@ -0,0 +1,482 @@
...
+        for (int i=0; i<k; i++)
+        {
+          poly x = J->m[i];
+          int li = getVariableIndex(x,currRing);
+          if (li>0)
+            satstdSaturatingVariables[i]=li;
+          else
+          {
+            WerrorS("satstd: second argument only ideals generated by variables supported for now");
+            return FALSE;
```
hannes14: should be: return TRUE;

YueRen:
done, thanks!

PR #763, opened 2016-04-28
reproducible: __DATE__ and __TIME__ : discard

jgmbenoit:
Attempt to render the upstream source reproducible:
https://wiki.debian.org/ReproducibleBuilds/TimestampsFromCPPMacros


hannes14:
Beside the official resleases we use many intermediate version of Singular,
and need the date to distinguish them, but I can offer:
8375099e2337c740a9324f313868ebd6de91ef7a

PR #766, opened 2016-05-19
L orderings over Z

adipopescu:


adipopescu:
i will merge it with spielwiese so there won't be a conflict anymore.

PR #768, opened 2016-05-20
Std over rings stuff

adipopescu:


hannes14:
resolved via manual merge

PR #776, opened 2016-07-31
throw error on too big or on negative shift

jakobkroeker:
currently a too big shift or negative shift will result in a segfault
(thus no hint about the cause)

=> throw an error with descriptive message.


hannes14:
right idea, wrong place: test user input in extra.cc,
see 462f5d0418503fbca8eddb710557f1e6e025ad6d

PR #778, opened 2016-08-02
chg: updated polymake interface to polymake 3.0

YueRen:
all examples now work with polymake 3.0 (and some only with polymake 3.0).

(old pull request had redundant changes in bbfan.cc, which have now been removed)


YueRen:
Hallo Hans, hab das polymake_wrapper mit deiner Version aktuallisiert, alle Beispiele gehen problemlos durch.

PR #789, opened 2016-09-20
Reduce bound

adipopescu:
add: system("reduce_bound",poly/vector/ideal/module, ideal/module, bound)

works like the normal form from global case (redNF) but over local/mixed ordering until the given bound.

Was properly tested by Gerhard


hannes14:
what is the difference of the new command system("...",p,I,B) to reduce(p,1,I,B) ?
see http://www.singular.uni-kl.de/Manual/4-0-3/sing_383.htm : reduce with 4 resp. 5 arguments

adipopescu:
This reduction doesn't consider the ecart although it can be used in the local and mixed case. 

If i use the normal reduction with 4,5 arguments, them it will go in redMora. If i use this one it goes to redNF even though the ordering is local.

PR #790, opened 2016-09-26
Tropical starting cone bug

YueRen:
Fixes a bug in the computation of starting cones and added various optimizations to make performance comparable to gfan.


hannes14:
Tst/Short/bug_newstruct.tst fails with:
 ? dynl_open failed:/tmp2/wawa/Tst/../Singular/MOD/gfanlib.so: undefined symbol: satstdSaturatingVariables
Probably LIB "customstd.so"; is required ?

YueRen:
Ah, yes. gfanlib.so now depends on customstd.so (for performance reasons), so everything which loads gfanlib.so now has to load customstd.so beforehand.

Changed the test with the latest commit.

PR #808, opened 2016-11-30
Allow libSingular and associated libraries to be built as shared libraries on Cygwin

embray:
See the original downstream issue here: https://trac.sagemath.org/ticket/21957

On Cygwin and MinGW, Singular's autotools files don't have the necessary bits to allow shared libraries to be built on Windows.  This is slightly fussier than on most *NIX, where a shared library may have unresolved symbols at link time.  On Windows, when linking a library as a DLL *all* symbols must be resolved at link time, which tends to expose areas where the build was (allowably) "inexplicit".


hannes14:
The patch does not work: build fails (in 32- and 64-bit version) because libomalloc is not build as a shared library.
(configure was: ./configure -C --prefix=....).
The original (unmodified) version builds fine ( a statically linked) Singular on the same machines

embray:
I'm not sure I fully understand--are you saying libomalloc *shouln't* be a shared lib, or that it just isn't?

hannes14:
If one links for example libfactory as a shared library, then all the the libs it depends on must also
be shared libraries (first example: libomalloc): but libomalloc is only build as static library
(have not tried to give additional argument to configure)
The same applies to libsingular_ressources and probably some more

embray:
I'm sorry, I'm still not following.  You wrote "the patch does not work", except it does work for me *on Cygwin*, insofar as it builds shared libraries (DLLs), which was my goal anyways.  Are you saying the goal is somehow wrong?  It doesn't seem to break anything on Linux either, but it may be that I don't know what I'm looking for.

hannes14:
For example, building a static version on linux fails:
configured with: 
{{{
./configure -C --enable-debug --disable-optimizationflags --enable-gfanlib --disable-shared --disable-p-procs-dynamic --enable-p-procs-static --with-python --with-builtinmodules=gfanlib,syzextra,customstd
}}}
build with: make
problem:
{{{
CXXLD    libSingular.la
libtool:   error: cannot find the library 'dyn_modules/gfanlib/gfanlib.la' or unhandled argument 'dyn_modules/gfanlib/gfanlib.la'
Makefile:1027: die Regel für Ziel „libSingular.la“ scheiterte
}}}
tried to build gfanlib.la seperatly:
{{{
cd Singular/dyn_modules/gfanlib
make
make: *** Keine Regel vorhanden, um das Ziel „../../../Singular/libSingular.la“, 
  benötigt von „gfanlib.la“, zu erstellen.  Schluss.
}}}
The problem here are dynamic modules (aka shared libraries) which can be loaded "on demand"
in a shared libary version of Singular (can can therefore build at any time, indep. of libSingular) but are part of the main executable for the static version and need to be compiled BEFORE libSingular
and nor depend on it for the static version: the problem can be solved by removing all dependencies
on libSingular from the Makefile in dyn_modules/gfanlib, but then the same occurs for syzextra
(the next module to build in).
As a dynamic module for Singular, they should not depend on libSingular anyway
(dynamic modules in the Singular definition are incomplete shared libaries, whose missing symbols
are in the main program)

embray:
> As a dynamic module for Singular, they should not depend on libSingular anyway
(dynamic modules in the Singular definition are incomplete shared libaries, whose missing symbols
are in the main program)

I understand that, but the problem is that on Windows shared libraries (i.e. DLLs) cannot be built unless all missing symbols can be resolved *at link time*.  This is in stark contrast to Linux where it's perfectly fine if not all symbols can be resolved at link time (for shared libs, though not for executables).

I see your point about building as a static lib though.  I think I have an idea for how to work around that.  Will update in a bit if it works.

hannes14:
One possibility would be to require on windows that these modules are part of the main program
(i.e. configure is called with --with-builtinmodules=....)

embray:
Picking back up on this work...

> One possibility would be to require on windows that these modules are part of the main program
(i.e. configure is called with --with-builtinmodules=....)

At the end of the day that might be simpler.  Though the solution I just pushed seems to work too.  It doesn't break anything on *nix as far as I can tell.  Both shared and static builds work there the same as before.  Instead, the tweaks needed for Windows are made only on Windows.  This works for modules listed in `--with-builtinmodules` as well.

embray:
Ah, I didn't notice you merged this.  Thanks!

PR #813, opened 2017-01-24
Fixed wrong call of surf execute script

sebasguts:
which caused Singular to return to main loop
too early and consequently not displaying the
created image in jupyter notebooks occasionly.


sebasguts:
Thank you.

PR #814, opened 2017-01-31
fix one error for sage in debug mode, see https://trac.sagemath.org/t…

jakobkroeker:
behebt eines der Probleme von 
https://trac.sagemath.org/ticket/21624;
("no minpoly allowed if there are local objects belonging to the basering")
siehe auch 
https://groups.google.com/forum/#!topic/libsingular-devel/tYuWlGOZdWo


jakobkroeker:
full example:
```
ring R = 0,(x,y,z,w),dp;
ideal I = x^3 + y^3 - z^3 - w^3, x^5 - y*z^4;
LIB("normal.lib");
genus(I);
```

PR #815, opened 2017-02-09
fix: plot in InteractiveShell

YueRen:
This will, in combination with Lars' latest updates in the InteractiveShell framework, create a popup with the image whenter 'plot' is called


YueRen:
Note: to identify whether Singular is running in a docker container inside InteractiveShell we check whether the current user is called "singularUser" using "whoami".

PR #816, opened 2017-02-15
Remove failing tests when SHIFTBBA is used.

jpflori:
This makes the warnings in the example from here:
* https://www.singular.uni-kl.de/Manual/4-0-2/sing_543.htm#SEC594
disappear.
I'm not sure this is correct though.
Please double check.


hannes14:
This would disable the tests allways - I think a better solution is 54c80cb81ed58f166a023fc060f541e24ebd5892
which usees strat->red==redFirstShift to check at runtime, if we are really called from bbaShift

PR #828, opened 2017-08-04
Added an uninterrupted build-in help browser

sebasguts:
For printing inline help in the jupyter kernel


hannes14:
A simpler solution provided by a9f04901d21b5556c3f290b250adadad68f10aa8,
and setting the global variable pagelength to -1

PR #829, opened 2017-08-14
rootisolation: tst files and various fixes

bendooru:
This introduces a test file for rootisolation.lib and the corresponding output.

In the process of creating tests various bugs were found and fixed, especially handling of edge cases was improved. The algorithm also checks its input for the necessary conditions and now accepts zero-dimensional ideals with an arbitrary number of generators.

Documentation and examples were added for C++ procedures that are available in the interpreter.

Indenting in .lib and .cc files was changed to 2 spaces as recommended by the style guide.

(If rootisolation.tst takes too much time to run the last example may be removed)


bendooru:
Auf Wunsch von Herrn Böhm vorerst geschlossen.

PR #833, opened 2017-11-06
truncated nchilb and right colon operation

shrawant:


hannes14:
merge manually.
Remarks: - test must not print system dep. data (like timings)
                - do not use cout/cerr: would be ignored by monitor et.

PR #836, opened 2017-11-12
rootisolation.tst

bendooru:
Tests functionality in rootisolation.lib and interval.so.

Includes some small fixes for interval.so.


hannes14:
example rootIsolationNoPreprocessing is different from provided rootisolation.res,gz.uu

bendooru:
Probably forgot `make install` before generating output. Fixed now.

PR #842, opened 2017-12-21
Updated versions of freegb.lib and fpadim.lib, new fpaprops.lib

kabouzeid:
These are the updated versions of freegb.lib and fpadim.lib
fpaprops.lib is a new library


hannes14:
The following tests fail (all from Tst/Manual):
'Example_of_use_of_@sc{Letterplace}.diff'   ivMis2Base.diff
 ivDHilbert.diff                            ivMis2Dim.diff
 ivDHilbertSickle.diff                      letplaceGBasis.diff
 ivDimCheck.diff                            lpNF.diff
 ivHilbert.diff                             makeLetterplaceRing.diff
 ivKDim.diff
Furthermore, lpNF uses >48GB of memory (instead of ~2.2 MB)

kabouzeid:
The tests you mentioned are working again

PR #848, opened 2018-02-01
templated call to pow() needs std::

dimpase:
Depending on C++ compiler and STL implementation, this is not guaranteed to work. In particular gcc 5.4 (or newer) on Solaris 11
errors out here. (In C++11 pow lives in namespace std, thus if this is not ensured then all the bets are off).

(normally, an instantiation of `int pow()` will  come from a template in STL, I think).

### DIFF
--- a/libpolys/coeffs/ffields.cc
+++ b/libpolys/coeffs/ffields.cc
@@ -806,7 +806,7 @@ static BOOLEAN nfCoeffIsEqual (const coeffs r, n_coeffType n, void * parameter)
 {
   if (n==n_GF) {
     GFInfo* p = (GFInfo *)(parameter);
-    int c = pow (p->GFChar, p->GFDegree);
+    int c = std::pow (p->GFChar, p->GFDegree);
     if ((c == r->m_nfCharQ) && (strcmp(n_ParameterNames(r)[0], p->GFPar_name) == 0))
       return TRUE;
   }
@@ -927,7 +927,7 @@ BOOLEAN nfInitChar(coeffs r,  void * parameter)
     return TRUE;
   }
 
-  int c = pow (p->GFChar, p->GFDegree);
+  int c = std::pow (p->GFChar, p->GFDegree);
 
   nfReadTable(c, r);
 


hannes14:
conversion to float/double for both arguments should make it indep. of C++11 or before

dimpase:
keeping it `int pow()` is safer from the point of view of possible overflows, IMHO.

hannes14:
pow(p->GFChar, p->GFDegree) cannot overflow as it is a number from a precomputed list
of tables and always <2^16
(must fit into unsigned short)

dimpase:
In such a case, using `double/float` to hold `int` looks to me like an overhead.

hannes14:
Had to revert this commit: fails to compile on FreeBSD with gcc 5.5.
And: it is an overhead, but it is a rarely used place.

dimpase:
FreeBSD has been using Clang as their standard compiler for years already,
building Singular just fine, why gcc 5.5?

I can look into it on FreeBSD 11, this is the only one I have access to atm.

hannes14:
clangs result also in an error(FreeBSD 11.1-RELEASE-p4, ver: 4.2.1 Compatible FreeBSD Clang 4.0.0 (tags/RELEASE_400/final 297347)):
ffields.cc:809:13: error: no member named 'pow' in namespace 'std'; did you mean
      simply 'pow'?
    int c = std::pow (p->GFChar, p->GFDegree);
            ^~~~~~~~
            pow
/usr/include/c++/v1/math.h:875:1: note: 'pow' declared here
pow(_A1 __lcpp_x, _A2 __lcpp_y) _NOEXCEPT

dimpase:
Could it be due to misnamed header?

By right it should be <cmath> rather than <math.h>

Needless to say it is all over Singular C++ code.
I have seen implementations of STL which are picky about this.

dimpase:
I've just tried the following on FreeBSD with clang 4.0.0:
```
#include <cmath>
int main()
{
int i,j,k;
i = 2;
j = 10;
k = pow(i,j);
return k;
}
```
And it works just fine. But as soon as I change `cmath` to `math.h` I get the error you report:
```
$ c++ xx.cc
xx.cc:7:5: error: no member named 'pow' in namespace 'std'; did you mean simply 'pow'?
k = std::pow(i,j);
    ^~~~~~~~
    pow
/usr/include/c++/v1/math.h:875:1: note: 'pow' declared here
pow(_A1 __lcpp_x, _A2 __lcpp_y) _NOEXCEPT
^
1 error generated.
```

Thus I very much suspect that the right thing to do is to change the Singular C++  `*.h` includes of `math.h`, `stdio.h`, and the like into `cmath`, `cstdio`, etc. And this will fix the error you report here, too.

PR #870, opened 2018-05-17
levandov: visual check for LP criterion in kutil and allowing module …

levandov:
…as input for freegb call. For testing purposes. Current problem: module in Letterplace setting is read as an ideal.

### DIFF
--- a/Singular/extra.cc
+++ b/Singular/extra.cc
@@ -1478,8 +1478,11 @@ BOOLEAN jjSYSTEM(leftv res, leftv args)
     if (strcmp(sys_cmd, "freegb") == 0)
     {
       const short t[]={3,IDEAL_CMD,INT_CMD,INT_CMD};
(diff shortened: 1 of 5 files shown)


Singular/extra.cc
```diff
@@ -1477,24 +1477,34 @@ BOOLEAN jjSYSTEM(leftv res, leftv args)
...
+      int uptodeg=(int)((long)(h->Data()));
+      h=h->next;
+      int lVblock=(int)((long)(h->Data()));
+      res->data = freegb(I,uptodeg,lVblock);
+      if (res->data == NULL)
+      {
+        /* that is there were input errors */
+        res->data = I;
+      }
+      if (iiCheckTypes(h,t,0))
```
hannes14: Here h is no longer the h from the beginning (see line 1489, 1491).
Better woould be not to change h (but to use a working pointer hh)
and to put here: res->rtyp=h->Typ();

PR #928, opened 2019-06-05
support strings representing general orderings in create_ring()

steenpass:


hannes14:
test differences in Tst/Manual/wurzel.tst

PR #930, opened 2019-06-27
add support for parameters givien as strings in create_ring()

steenpass:


fingolfin:
Typo in commit message: "givien"

PR #933, opened 2019-06-28
configure.ac: make libsingular-config executable

fingolfin:


hannes14:
Here libsingular-config has alreday the x-bit. But do it twice should not hurt.

fingolfin:
Was `libsingular-config` also executable if you removed it before running `configure`, resp. before running `./config.status libsingular-config`? Over here (both on macOS and Ubuntu), it is not.

And no surprise: that file is generated by piping the output of awk and sed into a file, and there is no reason why that should be executable unless one manually makes it so -- unless the file already exists and is writeable.

hannes14:
Yes. Situation before configure: libsingular-config did not exist in the source directory, nor in bin-install-dir.
After "configure;make install": libsingular-config was executable in bin-install-dir.

PR #934, opened 2019-06-28
Fix compiler warning by removing use of `register`

fingolfin:
For modern C compilers, this keyword has no effect at all (for GCC
at least since GCC 3, released almost two decades ago).

### DIFF
--- a/factory/ffops.h
+++ b/factory/ffops.h
@@ -151,7 +151,7 @@ inline int ff_inv ( const int a )
     if ( ff_big )
         return ff_biginv( a );
     else {
-        register int b;
+        int b;
         if ( (b = (int)(ff_invtab[a])) )
             return b;
         else


hannes14:
You are right for C++ compilers, but not for C compilers:
Timeline: 2011: C++11: register became deprecated, 2014: gcc 4.8 implements most of C++11,
2017: C++17: became reserved,  June 2018: clang on OsX starts with warnings about register

fingolfin:
I am not talking about depreciation, I am talking about effect: GCC 3 had a new register allocator which ignored the hint provided by the `register` keyword. AFAIK they never changed that, and with good reason.

So yeah, you can use it, but at least with gcc, it had no effect for the past 20 years

hannes14:
factory/ffops.h with register:
   text    data     bss     dec     hex filename
1777265   14448     256 1791969  1b57e1 libfactory-4.1.2.so
factory/ffops.h without register:
   text    data     bss     dec     hex filename
1777345   14448     256 1792049  1b5831 libfactory-4.1.2.so
other identical system (gcc 8, optimized with -O2).
The register keyword has beside being a hint for the allocation the effect
that the variable cannot be aliased - important for optimization.
Maybe this shows up here.

PR #935, opened 2019-06-30
Remove some useless files

fingolfin:
The files AUTHORS, ChangeLog and NEWS were empty, let's remove them
(they probably were there because automake complains if they are
missing -- unless one passes the `foreign` option to it, which we do).

The file clone_autogen_generated_stuff.sh clones a 6 year old repository,
apparently in an effort to not make the user need autotools installed.
Which seems like a bad idea to start with, but it also hasn't been kept
up-to-date. So let's just ditch it.

I would also recommend to delete the repository https://github.com/Singular/AUTOGEN_generated_stuff or at least archive it.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,6 +4,4 @@
 /doc/ export-ignore
 /factory/old/ export-ignore
 /old_modgen/ export-ignore
-/clone_autogen_generated_stuff.sh export-ignore
 /git-archive export-ignore
-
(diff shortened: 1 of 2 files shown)


hannes14:
agreed for clone_autogen_generated_stuff.sh (and gitattributes): we try to simplify/sped up the autoconf/automake stuff - but is is difficult to maintain and introduces more dependcies.
But we should keep AUTHORS, ChangeLog and NEWS as automake wants them.

fingolfin:
Are you sure automake wants them? As explained above, it does not want them over here, due to the `foreign` option

PR #938, opened 2019-07-05
Parallel flint conversion

tthsqe12:
there is a switch that needs a proper condition to detect if omalloc (or any other non-threadsafe allocator) is being used.


libpolys/polys/flint_mpoly.cc
```diff
@@ -36,31 +40,51 @@ BOOLEAN convSingRFlintR(fmpq_mpoly_ctx_t ctx, const ring r)
...
+  {
+    nmod_mpoly_ctx_init(ctx,r->N,ORD_LEX,r->cf->ch);
+    return FALSE;
+  }
+  return TRUE;
+}
+
+/******** polynomial conversion ***********/
+
+#if 1
```
tthsqe12: This is the line that needs to change.

Issue #925, opened 2019-05-17
Performance of `std` vs. `liftstd`


zickgraf:
I want to compute the transformation matrix from a given module over a super-commutative ring to a standard basis of this module. My current approach is to use `liftstd` (see code below). I expect the runtime of `liftstd` to be similar to the runtime of `std`, because in the algorithms I know, `liftstd` is "`std` + some bookkeeping". This is correct in some cases, but in other cases, `std` finishes in one second, whereas `liftstd` is still running after some minutes (I did not check if it terminates eventually).

Is this behaviour expected or should this be considered as a bug?

If this is a bug, I will provide a concrete example, of course, but since my example is large, I do not want to clutter this issue right now.

```
LIB "nctools.lib";

ring R = (0),(e0,e1,e2),dp;
def S = superCommutative(1);
setring S;

matrix M[702][216] = ...; // large matrix

matrix T;
module s;
std(module(M)); // terminates after one second
liftstd(module(M),T,s,"std"); // still running after some minutes
```

zickgraf:
I have investigated this a bit more: the difference in performance also occurs for commutative rings. In one example, `std` finished in less than a second but `liftstd` takes nearly three minutes.

hannes14:
solved with b88827821a3ed7373494ab9c988f91d03cfc2f68

PR #965, opened 2019-11-27
Improve performance of liftstd

zickgraf:
I'm investigating the performance of `liftstd` (see #925) and think there is a large potential for optimizations. This commit stops the reduction in `redHoney` once the entries of the actual Gröbner basis are zero (instead of continuing the reduction with the entries of the transformation matrix). This is analogous to https://github.com/zickgraf/Singular/blob/0bb946880d425f56c1db84b42ea61b5ece5de8e9/kernel/GBEngine/kstd2.cc#L1782 but I do not delete `h` and return 1 instead of 0 since the nonzero part of `h` is needed for the syzygies. This improves the runtime of `liftstd` in the following example from 6900ms to 1700ms on my computer (for larger examples the difference is even greater):

```
LIB "random.lib";
// make random reproducible
system("random", 1);
system("--ticks-per-sec",1000);

int t;

ring R = 0,(x,y,z),(c,dp);
matrix A = randommat(20, 700, maxideal(1), 10);
// we want to test redHoney, so A should not be homogeneous since in this case redHomog would be used
A[1,1] = x^2;

t = timer;
matrix G1 = std(A);
print("computed std in " + string(timer - t) + " ms");

matrix T;
module S;
t = timer;
matrix G2 = liftstd(A, T, S);
print("computed liftstd in " + string(timer - t) + " ms");
```

If this PR gets merged I will provide PRs for the other `red...` functions.

Edit: Of course this changes the result of syzygy computations so some tests are failing. If we are sure that this change is mathematically correct, I will update the tests.

### DIFF
--- a/kernel/GBEngine/kstd2.cc
+++ b/kernel/GBEngine/kstd2.cc
@@ -1798,6 +1798,23 @@ int redHoney (LObject* h, kStrategy strat)
         }
       }
     }
+    if (strat->syzComp > 0)
+    {
+      if (h->p!=NULL)
+      {
+        if(p_GetComp(h->p,currRing)>strat->syzComp)
+        {
+          return 1;
+        }
+      }
+      else if (h->t_p!=NULL)
+      {
+        if(p_GetComp(h->t_p,strat->tailRing)>strat->syzComp)
+        {
+          return 1;
+        }
+      }
+    }
     h->SetShortExpVector();
     not_sev = ~ h->sev;
     h_d = h->SetpFDeg();


zickgraf:
Thanks for adjusting the tests in 0bb2274617800aa224288b50df65899ef7420528!

I have two follow-up questions:
1. Why is it necessary to disable the optimization for `syz`? The test you mention in the commit message (Tst/New/modulo_homalg_full.tst) still runs fine with the optimization enabled on my computer, and the documentation of `syz` says, if `returnSB` is not set, a generating set of the syzygy module is returned, which I think should still hold true with the optimization enabled. Am I missing something?
2. I took another look at the existing optimization, which I borrowed from here:
https://github.com/zickgraf/Singular/blob/0bb946880d425f56c1db84b42ea61b5ece5de8e9/kernel/GBEngine/kstd2.cc#L1782 
This optimization seems to be enabled here: https://github.com/zickgraf/Singular/blob/0bb946880d425f56c1db84b42ea61b5ece5de8e9/kernel/ideals.cc#L911 If I understand the code correctly, `k` is the rank of the module. Why is the optimization not enabled for modules with rank > 1?

fingolfin:
@hannes14 could you comment on @zickgraf's questions?

hannes14:
to 2: see for example Tst/Old/liftstan.tst (with ring r1 = 32003,(x,y,z),(c,ls);):
the GB of [[x4,y2,xz2].[y3,x8,2z4] is  [y3,x8,2z4][x4,y2,xz2][0,y5-x12,xy3z2-2x4z4],
while liftstd without k==1-test gives: [y3,x8,2z4][x4,y2,xz2].
This is probably a wrong check for being a GB-element or a syzygy (>1 vs >syzComp?)

zickgraf:
> This is probably a wrong check for being a GB-element or a syzygy (>1 vs >syzComp?)

Do I understand you correctly, that this wrong check is a bug somewhere else in the code?

I think my change basically replicates the existing `TEST_OPT_IDLIFT` optimization in a slightly different way. So if that bug was fixed we could probably revert my change again (I would, of course, test this before reverting).

PR #1004, opened 2020-06-17
Fix an automake warning

fingolfin:


Singular/dyn_modules/gfanlib/Makefile.am
```diff
@@ -30,7 +30,7 @@ gfanlib_la_CPPFLAGS = ${MY_CPPFLAGS} ${P_PROCS_CPPFLAGS_COMMON} ${CDDGMPCPPFLAGS
 gfanlib_la_LDFLAGS  = ${P_PROCS_MODULE_LDFLAGS} $(SINGULAR_LDFLAGS)
 
 # forcefully enable exceptions for polymake
-CXXFLAGS+= $(FEXCEPTIONSFRTTI_CXXFLAGS)
+AM_CXXFLAGS = $(FEXCEPTIONSFRTTI_CXXFLAGS)
```
hannes14, line 34: Does not work:
.../Singular/dyn_modules/gfanlib/ppinitialReduction.cc: In Funktion »void pReduce(spolyrec*&, number, ring)«:
.../Singular/dyn_modules/gfanlib/ppinitialReduction.cc:93:19: Fehler: Ausnahmebehandlung ausgeschaltet, »-fexceptions« schaltet sie ein
   93 |             throw 0;
      |                   ^
make[5]: *** [Makefile:692: gfanlib_la-ppinitialReduction.lo] Fehler 1

We need to overwrite the general -fno-exceptions for this, which CXXFLAGS does,
 but AM_CXXFLAGS are prepended, so  -fexceptions has no effect here.
The same applies also to the other CXXFLAGS  changes below.

fingolfin: OK, I guess the reasons I didn't see those issues locally is because gfanlib is not built for me (which I didn't notice before. Ooops. Well, I am kinda used to submitting PRs and then having CI comment to point out these things for me... Luckily I have you, Hans, thank you :-).

I'll drop these then for now. But really, the proper change then likely is to not have `-fno-exceptions` in the global CXXFLAGS either, and instead only add it in places that want it. Hm...

PR #1013, opened 2020-07-03
Fix a bunch of warnings

fingolfin:
Please somebody verify I didn't screw anything up by "fixing" it (I tried to be diligent, but...)

I think a few of them fix actual (hopefully harmless) bugs

### DIFF
--- a/Singular/dyn_modules/Order/nforder_ideal.cc
+++ b/Singular/dyn_modules/Order/nforder_ideal.cc
@@ -310,6 +310,7 @@ nforder_ideal* nf_idPower(nforder_ideal* A, int i)
     return new nforder_ideal(A, 1);
   } else if (i<0) {
     Werror("not done yet");
+    return NULL;
   } else {
     nforder_ideal *B = nf_idPower(A, i/2);
     nforder_ideal *res = nf_idMult(B, B);
--- a/Singular/dyn_modules/cohomo/cohomo.cc
+++ b/Singular/dyn_modules/cohomo/cohomo.cc
@@ -1312,9 +1312,10 @@ std::vector<int> fvarsvalue(int vnum, std::vector<int> fvars)
     if(fset[i]==vnum)
     {
       fset.erase(fset.begin()+i);
-      return fset;
+      break;
     }
   }
+  return fset;
 }
 
 
@@ -4134,7 +4135,7 @@ std::vector<std::vector<int> > bsubsets_1(poly b)
   std::vector<std::vector<int> > bset;
   for(int i=0;i<bvs.size();i++)
   {
-    for(int j=0;j<bvs.size(), j!=i; j++)
+    for(int j=0;j!=i; j++)
     {
       vs.push_back(bvs[j]);
     }
--- a/Singular/dyn_modules/systhreads/libthread.h
+++ b/Singular/dyn_modules/systhreads/libthread.h
@@ -24,7 +24,7 @@ namespace LibThread {
   // Job scheduling
   Job *startJob(ThreadPool *pool, Job *job, leftv arg);
   Job *startJob(ThreadPool *pool, Job *job);
-  Job *scheduleJob(ThreadPool *pool, Job *job, long ndeps, Job **deps);
+//   Job *scheduleJob(ThreadPool *pool, Job *job, long ndeps, Job **deps);
   void cancelJob(Job *job);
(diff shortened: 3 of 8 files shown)


Singular/dyn_modules/Order/nforder_ideal.cc
```diff
@@ -310,6 +310,7 @@ nforder_ideal* nf_idPower(nforder_ideal* A, int i)
     return new nforder_ideal(A, 1);
   } else if (i<0) {
     Werror("not done yet");
+    return NULL;
```
fingolfin, line 313: Looking at `Werror`, it seems it actually *can* return?!? If so, this return is needed. If I am mistaken, perhaps a `noreturn` annotation could be added to it and this `return` be removed again.

Singular/dyn_modules/cohomo/cohomo.cc
```diff
@@ -4134,7 +4135,7 @@ std::vector<std::vector<int> > bsubsets_1(poly b)
   std::vector<std::vector<int> > bset;
   for(int i=0;i<bvs.size();i++)
   {
-    for(int j=0;j<bvs.size(), j!=i; j++)
+    for(int j=0;j!=i; j++)
```
fingolfin, line 4138: The new code is equivalent to the old, and also seems sensible to me, but somebody who knows this code should verify this is what was the intention (and that not e.g. `&&` instead of `,` was intended)

Singular/dyn_modules/systhreads/libthread.h
```diff
@@ -24,7 +24,7 @@ namespace LibThread {
   // Job scheduling
   Job *startJob(ThreadPool *pool, Job *job, leftv arg);
   Job *startJob(ThreadPool *pool, Job *job);
-  Job *scheduleJob(ThreadPool *pool, Job *job, long ndeps, Job **deps);
+//   Job *scheduleJob(ThreadPool *pool, Job *job, long ndeps, Job **deps);
```
fingolfin, line 27: grep didn't find any caller for this (@rbehrends or did I miss something?)

Singular/dyn_modules/systhreads/shared.cc
```diff
@@ -1479,6 +1479,7 @@ void *joinThread(ThreadState *ts) {
   ts->running = false;
   ts->active = false;
   thread_lock.unlock();
+  return result;
```
fingolfin, line 1482: Nothing ever seems to use the return value, so there wasn't a problem here. So an alternative fix would be to remove the return value altogether, but I reckon it is intended for a future usecase? @rbehrends ?

Singular/svd_si.h
```diff
@@ -3224,7 +3224,7 @@ namespace bidiagonal
         {
             return;
         }
-        ap::ap_error::make_assertion(fromtheright && zcolumns==m || !fromtheright && zrows==m);
+        ap::ap_error::make_assertion(fromtheright ? zcolumns==m : zrows==m);
```
fingolfin, line 3227: Note that `||` bind stronger than `&&` binds stronger than `==`, so the old code checked `(fromtheright && zcolumns)== ((m || !fromtheright) && zrows) ==m`

Singular/dyn_modules/systhreads/libthread.h
rbehrends: This exists as a C++ API version of the Singular function `scheduleJob()`, which is implemented independently. It is not called by the implementation of the Singular function, because validation and error checking is interleaved with the actual logic. This is some unfortunate code duplication that should be cleaned up at some point.

Singular/dyn_modules/systhreads/shared.cc
rbehrends: This is part of the public C++ API (see the still incomplete `doc/kernel.md`), so it should indeed return `result`.

Singular/dyn_modules/Order/nforder_ideal.cc
hannes14: Werror always returns: usually it write the error message and sets a flag (inerror) which is used by the interpreter:
the interpreter skips the curent result, returns to the top level and resets the flag.

fingolfin: OK in that case I think my change is correct.

PR #1015, opened 2020-07-05
Move FLINTrandom into factory to fix test suite crashes

fingolfin:
Without this, `make check` runs into segfaults in tests that use factory but
not the full Singular library, as factory accesses FLINTrandom but does not
contain its definition.


fingolfin:
Note that this fixes a failure I experienced on macOS when using flint but not NTL. But as a matter of fact, when using NTL but not flint, I also get a segfault in factory's `make check` on Linux. I have not yet tried to figure out what causes that.

PR #1019, opened 2020-07-08
Add basic GitHub Actions CI tests

fingolfin:
This adds basic CI tests to this repository using GitHub Actions. For now, all these do is to build Singular and then run `make check`, with a build matrix consisting of four jobs: 
- using Ubuntu or macOS
- using a default configure invocation, or else one with `--without-ntl --with-flint`

Note that on macOS, Flint 2.6.0 is installed, while on Ubuntu Flint 2.5.2. As a result, one of the four builds (the one using `--with-flint` on the Ubuntu machine with Flint 2.5.2) fails (due to a segfault in factory's `make check). I left it in as commit 249d818e57a3c49f4e561e2a5fe38db7dbc9f2c0 indicates that Flint 2.5.2 ought to be supported. I guess we should either fix `make check` there or else change configure to require Flint 2.6?

### DIFF
--- a/.github/workflows/runtests.yml
+++ b/.github/workflows/runtests.yml
@@ -1,0 +1,42 @@
+name: Run tests
+
+# Trigger the workflow on push or pull request
+on: [push, pull_request]
+
+env:
+  prefix: "/tmp/prefix"
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest]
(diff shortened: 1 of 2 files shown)


hannes14:
make check for factory now also works with FINT 2.5.2 (and without NTL)
make dist is used (in the form of make distcheck) during release to test makefiles.
make dist should now also work (with and without doc/doc.tbz2)

fingolfin:
Thanks @hannes14, indeed all tests pass now, see https://github.com/fingolfin/Singular/actions/runs/162972872

fingolfin:
This is ready for merging (from my point of view at least :-)

PR #1021, opened 2020-07-09
More CI tests on GitHub action

fingolfin:
This is based on PR #1019 and adds more tests, which for now however mostly are failing:
- on Ubuntu when building with NTL but without Flint, `cd Tst && ./regress.cmd -s ../Singular/Singular Old/universal.lst` produces this failure (interestingly, in the macOS build this works fine):
```
--- proc 
64d63
< // flintQ                         [0]  proc from kernel (C)
101d99
!!! proc : Differences in res files
< // flintQ                         [0]  proc from kernel (C)
143d140
< // flintQ                         [0]  proc from kernel (C)
```
 - on Ubuntu when building using FLINT but not NTL, some tests fail because of different results; e.g. "multivariate factorization over Z depends on NTL(missing)" is shown in the `factor` test

There's more but those are the first two that I run into

### DIFF
--- a/.github/workflows/runtests.yml
+++ b/.github/workflows/runtests.yml
@@ -12,23 +12,29 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
-        configflags: ["", "--without-ntl --with-flint"]
+        configflags: ["", "--with-ntl --with-flint", "--without-ntl --with-flint"]
 
     steps:
       - uses: actions/checkout@v2
       - name: "Install dependencies"
         run: |
+               FLINT_JLL="https://github.com/JuliaBinaryWrappers/FLINT_jll.jl/releases/download/FLINT-v200.700.0%2B0/FLINT.v200.700.0"
                if [ "$RUNNER_OS" == "Linux" ]; then
                     # sharutils is for uudecode
                     sudo apt install sharutils libgmp-dev libreadline-dev libmpfr-dev libntl-dev libcdd-dev 4ti2 normaliz
-                    # install new enough FLINT (>= 2.6.0)
-                    wget -O FLINT.tar.gz "https://github.com/JuliaBinaryWrappers/FLINT_jll.jl/releases/download/FLINT-v2.6.0%2B0/FLINT.v2.6.0.x86_64-linux-gnu.tar.gz"
+                    # install new enough FLINT
+                    wget -O FLINT.tar.gz "${FLINT_JLL}.x86_64-linux-gnu.tar.gz"
                     sudo tar -C /usr -xvf FLINT.tar.gz
                     rm -f FLINT.tar.gz
                elif [ "$RUNNER_OS" == "macOS" ]; then
-                    brew install autoconf automake libtool gmp readline mpfr ntl flint cddlib
+                    brew install autoconf automake libtool gmp readline mpfr ntl cddlib
                     # TODO: 4ti2?
                     # TODO: normaliz?
+                    # install new enough FLINT
+                    wget -O FLINT.tar.gz "${FLINT_JLL}.x86_64-apple-darwin.tar.gz"
+                    sudo mkdir -p /usr/local
+                    sudo tar -C /usr/local -xvf FLINT.tar.gz
+                    rm -f FLINT.tar.gz
                else
                     echo "$RUNNER_OS not supported"
                     exit 1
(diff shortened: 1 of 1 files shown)


hannes14:
flintQ  is a procedure which allows for flints fmpq_mpoly as coefficients of Singular polynomials
- it is clearly not available if Singular was compiled without Flint.
Other difference which may occur: different use of the random generator, resulting in
different order (of factors) or different signs of generators of ideals.
The pre-recorded test results expect: Singular with NTL ,FLINT 2.6, 4ti2, normaliz in the path.
For these tests one can collect the of all this indepent tests into a seperate .lst file and use regress.cmd with it

fingolfin:
@hannes14 some tests fail with messages like `? cannot open customstd.so`or `? cannot open gfanlib.so`, even though these have been built. Is it possible that they are only found (and thus can only be loaded) if Singular was installed via `make install`? If so, I guess I could introduce a step doing just that

fingolfin:
I changed this to do `make install`, and also installed Flint 2.6.0 on the Linux machine, too (with Flint 2.5.2, several things are not implemented, leading to many additional failures in the test suite).

hannes14:
Yes, if dynamic modules are used, "make install" is required.

fingolfin:
OK, getting closer there; disabled some tests and now it passes at least when using NTL. One test I had to pause as it run for hours, even with NTL, was `Long/factorizeQ_l`:
```
Thu, 09 Jul 2020 23:39:48 GMT  --- factorizep_l            15.30 
Fri, 10 Jul 2020 03:01:17 GMT  Killed
Fri, 10 Jul 2020 03:01:17 GMT  !!! factorizeQ_l : Singular call exited with status (35072) != 0
Fri, 10 Jul 2020 03:01:17 GMT  --- factorizeQ_l 
Fri, 10 Jul 2020 03:02:13 GMT  --- factorizeQa_l           55.32 
```
Any idea what might be going on there?


When using FLINT instead of NTL, things are worse, simply because the missing multivariate factorization pops up in quite a few tests. So I would try to just disable those tests in that combo for now, and reactivate them once this feature is implemented.

hannes14:
On Fri, Jul 10, 2020 at 12:54:09AM -0700, Max Horn wrote:
> OK, getting closer there; disabled some tests and now it passes at least when using NTL. One test I had to pause as it run for hours, even with NTL, was `Long/factorizeQ_l`:

This is a known bug in FLINT 2.6.0
> 
> When using FLINT instead of NTL, things are worse, simply because the missing multivariate factorization pops up in quite a few tests. So I would try to just disable those tests in that combo for now, and reactivate them once this feature is implemented.

Still in process of implementing stuff when using FLINT instead of NTL.

fingolfin:
> On Fri, Jul 10, 2020 at 12:54:09AM -0700, Max Horn wrote: OK, getting closer there; disabled some tests and now it passes at least when using NTL. One test I had to pause as it run for hours, even with NTL, was `Long/factorizeQ_l`:
> This is a known bug in FLINT 2.6.0

Excellent. Is it already fixed in FLINT master? If not, can you point me at the issue on the issue tracker?

hannes14:
https://github.com/wbhart/flint2/issues/783

fingolfin:
@hannes14 there are quite a few test failures now even when building with NTL and without FLINT. (CC @mohamed-barakat @tthsqe12). See e.g. https://github.com/Singular/Singular/pull/1021/checks?check_run_id=1298654743

tthsqe12:
I am not too familiar with Singular's testing strategy, but I believe the comparing of textual output makes it a bit more fragile than in needs to be. Most (all?) of the function provided by factory/flint/ntl are uniquely defined with the exception of factoring. 
I can see two routes:
1. The singular test code probably tests functions whose output is not easily verifiable, and thus must be compared with known correct results. If the correct answer is, say, `ideal<x,y,z>`, instead of printing the answer via `func();` the test code could do something like `@test func() == ideal<x,y,z>` or a cheaper version of this. This would be a lot of work to change all of the test code.
2. On several occasions when working over finite fields, I have noticed that singular (with factory) arbitrarily outputs non-monic factors. This is not nice. The only reason being for this is that the factors were monic in factory's lex ordering, but became non-monic when converted to the actual ring ordering. Therefore some normalization of the factors could happen in singclap_factorize, such the the result of singclap_factorize is uniquely defined. The test code verification files would just have to be recompiled once.

fingolfin:
I've updated this PR to use the latest FLINT_jll, which is based on a current snapshot of the FLINT repository.

I still see tests failing with error `multivariate factorization over Q(alpha) depends on NTL(missing)`. But we already established that NTL doesn't implement this factorization; so shouldn't it be possible to adapt the function `ratFactorize` in Singular to also work with FLINT?

tthsqe12:
My next project is figuring out to call `Hecke.factor(::Hecke.Generic.MPoly{nf_elem})` from c/c++ code.

fingolfin:
It seems the new tests are now all passing! Great work, @hannes14 and @tthsqe12 ! I've just rebased and squashed the changes; assuming they all pass (ignoring the "Sage" ones), I hope this can be finally merged.

PR #1026, opened 2020-08-18
autoconf: improve cddlib and polymake detection

fingolfin:
For the cddlib detection (used for gfanlib), switch to using AC_CHECK_LIB;
this gives better diagnostics, and since it is "standard", is a bit easier to
debug.

For polymake, when checking for `polymake-config` via `AC_CHECK_PROG`, use
`yes` and `no` instead of `1` and `0` as results. This let's the configure
script print nicer messages ("checking for FOO... yes" vs "checking for FOO...
1")


hannes14:
libcdd may be present, but the include files not(debian). In this case gfanlib can not be build.

fingolfin:
Updated to correctly handle the case of missing `setoper.h`.

PR #1032, opened 2020-11-26
Remove old_modgen, modules, factory/old

### LINKED ISSUE #687: some (deprecated?) files containes deprecated SINGUNAME
some (deprecated?) files containes deprecated SINGUNAME

(grep for it)

fingolfin:
Resolves #687

Even if there is something in there worth preserving, it *is* preserved: in the git history.


hannes14:
without factory/old/gengftables.cc factory/gftables/* cannot be regenerated.
This file should be kept (possible moved to factory/.)

PR #1037, opened 2021-01-09
Change ZnmInfo member exp to unsigned long

fingolfin:
... instead of int. This is the type used everywhere else for the exponent,
e.g. for modExponent

### DIFF
--- a/libpolys/coeffs/rmodulon.h
+++ b/libpolys/coeffs/rmodulon.h
@@ -15,7 +15,7 @@ struct snumber; typedef struct snumber *   number;
 #ifdef HAVE_RINGS
 #include "coeffs/rintegers.h"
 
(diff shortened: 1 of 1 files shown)


fingolfin:
OK this doesn't build because of the use of `nEati` to deserialize. I can fix that, of course, but before I spend effort on that, it'd be good to know if this PR is even in the right direction

hannes14:
Am Mon, Jan 11, 2021 at 07:52:48AM -0800 schrieb Max Horn:
> OK this doesn't build because of the use of `nEati` to deserialize. I can fix that, of course, but before I spend effort on that, it'd be good to know if this PR is even in the right direction

Fine with me, but then one needs an additional routine nEatl (like
nEati).
On the other side - is "unsigned long" needed here?

hannes14:
merged manually

PR #1038, opened 2021-01-09
[prototype] Improve performance of liftstd

zickgraf:
This PR drastically improves the performance of `liftstd` in cases where a lot of syzygies appear during the reductions.

Consider the following example:
```
LIB "random.lib";
// make random reproducible
system("random", 1);
system("--ticks-per-sec",1000);

int t;

ring R = 0,(x,y,z),(c,dp);
matrix A = randommat(30, 700, maxideal(1), 10);
// use redHoney (so A should not be homogeneous since in this case redHomog would be used)
A[1,1] = x^2;

t = timer;
matrix G1 = std(A);
print("computed std in " + string(timer - t) + " ms");

matrix T;
t = timer;
matrix G2 = liftstd(A, T);
print("computed liftstd in " + string(timer - t) + " ms");
```

Without the PR, I get the following runtimes:
```
computed std in 2300 ms
computed liftstd in 9910 ms
```
so `liftstd` is slower than `std` by more than a factor of 4.

With the PR, I get the following runtimes (of course I have verified that I get the same Groebner basis `G2` and transformation matrix `T`):
```
computed std in 2400 ms
computed liftstd in 2810 ms
```
so `liftstd` only takes 17% longer than `std`. 

If I create `A` with more than 30 rows, the relative improvement gets even larger. In particular, `liftstd` now has the runtime I expected in #925, namely the runtime of "`std` + some bookkeeping".

# Why is `liftstd` slow in this example?

The columns of the input matrix `A` define polynomials/vectors `p_i`. We want to compute a Groebner basis `G` of the ideal/module generated by the `p_i`, together with a transformation matrix `T` such that `A * T == G`. To this end, `liftstd` puts an identity matrix below `A` to which the standard Groebner basis algorithm is applied (with some optimizations to not compute a Groebner basis below the lines corresponding to `A`). So the standard Groebner basis algorithm now does the reductions with very long vectors: the original vectors `p_i` + the transformation part, which is the sparse identity matrix at the beginning of the alorithm but quickly fills up. Of course, this is expected at first: if we want to compute the transformation matrix, we have to deal with long vectors. However, this approach also needlessly computes all the syzygies: It does the reductions with the long vectors, and if it sees that the components of the result corresponding to the actual Groebner basis are 0 (i.e., the transformation part is a syzygy), it throws the result away. Thus, it is much more efficient to compute the transformation part lazily: do the reductions with the short vectors, record the coefficients used during the reduction, and compute the transformation part of the result if (and only if) we do not have a syzygy.

# Details of the code of the PR

1. In `bba`, before a polynomial `P` (in the following called the dividend) is reduced, its transformation part (which might be distributed among the various buckets) is extracted and stored in `P.transformation_coeffs`.
2. Now, every time the dividend `PR` is reduced by a divisor `PW` via `P = LC(PW)*PR - LT(PR)/LM(PW)*PW` in `ksReducePoly` do the following:
    1. Extract the transformation part of `PW` and store it in `PW.transformation_coeffs`.
    2. Compute the reduced polynomial `P` as before.
    3. Add the number `LC(PW)` and the polynomials `LT(PR)/LM(PW)` and `PW.transformation_coeffs` to the arrays `PR->transformation_coeffs_parts_numbers`, `PR->transformation_coeffs_parts_lms` and `PR->transformation_coeffs_parts_divisor_transformation_coeffs`.
    4. Reattach the transformation part of the divisor stored in `PW.transformation_coeffs` to `PW`.
3. If after the reductions we do not have a syzygy: compute the transformation part of the result using the information stored in `P.transformation_coeffs_parts_numbers`, `P.transformation_coeffs_parts_lms` and `P.transformation_coeffs_parts_divisor_transformation_coeffs` and attach it to the result.

# State of the PR

This PR is only a prototype: it has many unhandled cases (some being caught explicitly in the code) and is only implemented for `bba` and `ksReducePoly`. Also, one could probably improve the performance even more by keeping the transformation part separate during the whole algorithm instead of extracting it before the reduction and reattaching it afterwards.

Currently, I do not need this improvement anymore (as I have moved the bottlenecks in my computations away from `liftstd`) and I certainly lack the knowledge of Singular to implement all of this properly. But if anyone with a deeper knowledge of Singular is interested in this, I would be happy to work together to actually get this working beyond a prototype :-)

### DIFF
--- a/kernel/GBEngine/kspoly.cc
+++ b/kernel/GBEngine/kspoly.cc
@@ -188,6 +188,51 @@ int ksReducePoly(LObject* PR,
                  number *coef,
                  kStrategy strat)
 {
+  if(strat != NULL && strat->syzComp > 0 && !(TEST_OPT_REDTAIL_SYZ)) {
+  
+      
+      // extract transformation coeffs of divisor
+      if(PW->p != NULL) {
+          
+          poly polynomial = PW->p;
+          
+          while (pNext(polynomial) != NULL)
+          {
+              if(pGetComp(pNext(polynomial)) > strat->syzComp) {
+                  
+                  if(PW->transformation_coeffs != NULL && PW->transformation_coeffs != pNext(polynomial)) {
+                      printf("transformation_coeffs:\n");
+                      pWrite(PW->transformation_coeffs);
+                      printf("polynomial:\n");
+                      pWrite(pNext(polynomial));
+                      printf("got transformation coeffs from somewhere unexpected\n");
+                      exit(1);
+                  }
+  
+                  PW->transformation_coeffs = pNext(polynomial);
+                  pNext(polynomial) = NULL;
+                  
+                  if(PW->transformation_coeffs_length == 0) {
+                      PW->transformation_coeffs_length = pLength(PW->transformation_coeffs);
+                  }
+  
+                  PW->pLength -= PW->transformation_coeffs_length;
+                  
+                  break;
+              }
+              pIter(polynomial);
+          }
+      }
+      else {
+          printf("PW->p is NULL, reducing by zero?\n");
+          exit(1);
+      }
+      if(PW->transformation_coeffs == NULL && strat != NULL && strat->syzComp > 0) {
+          printf("could not find transformation coeffs of divisor\n");
+          exit(1);
+      }
+  }
+
 #ifdef KDEBUG
   red_count++;
 #ifdef TEST_OPT_DEBUG_RED
@@ -238,10 +283,12 @@ int ksReducePoly(LObject* PR,
   }
 #endif
 
-  if (t2==NULL)           // Divisor is just one term, therefore it will
+  if (t2==NULL && false)           // Divisor is just one term, therefore it will
   {                       // just cancel the leading term
     PR->LmDeleteAndIter();
     if (coef != NULL) *coef = n_Init(1, tailRing->cf);
+    printf("unhandled case: divisor is only a single term\n");
+    exit(1);
     return 0;
   }
 
@@ -254,8 +301,16 @@ int ksReducePoly(LObject* PR,
     {
       // undo changes of lm
       p_ExpVectorAdd(lm, p2, tailRing);
-      if (strat == NULL) return 2;
-      if (! kStratChangeTailRing(strat, PR, PW)) return -1;
+      if (strat == NULL) {
+          printf("unhandled case 3\n");
+          exit(1);
+          return 2;
+      }
+      if (! kStratChangeTailRing(strat, PR, PW)) {
+          printf("unhandled case 4\n");
+          exit(1);
+          return -1;
+      }
       tailRing = strat->tailRing;
       p1 = PR->GetLmTailRing();
       p2 = PW->GetLmTailRing();
@@ -277,16 +332,23 @@ int ksReducePoly(LObject* PR,
 #endif
 
   // take care of coef buisness
+  if(PR->transformation_coeffs != NULL) {
+    (*PR->transformation_coeffs_parts_numbers)[PR->transformation_coeffs_parts_length] = NULL;
+  }
   if (! n_IsOne(pGetCoeff(p2), tailRing->cf))
   {
     number bn = pGetCoeff(lm);
     number an = pGetCoeff(p2);
     int ct = ksCheckCoeff(&an, &bn, tailRing->cf);    // Calculate special LC
     p_SetCoeff(lm, bn, tailRing);
-    if ((ct == 0) || (ct == 2))
+    if ((ct == 0) || (ct == 2)) {
+      if(PR->transformation_coeffs != NULL) {
+          (*PR->transformation_coeffs_parts_numbers)[PR->transformation_coeffs_parts_length] = an;
+      }
       PR->Tail_Mult_nn(an);
+    }
     if (coef != NULL) *coef = an;
-    else n_Delete(&an, tailRing->cf);
+    if(coef == NULL && (PR->transformation_coeffs == NULL || (*PR->transformation_coeffs_parts_numbers)[PR->transformation_coeffs_parts_length] == NULL)) n_Delete(&an, tailRing->cf);
   }
   else
   {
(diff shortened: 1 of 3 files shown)


fingolfin:
This sounds pretty cool. I wonder if @hannes14 has any thoughts on this?

hannes14:
Impressive and very interessting - I appriciate the ideas behind this patch,
but have problems with the real patch:
- it nearly doubles the size of TObject (nad LObject) in best case, increase the size in the worst case by 16 kB (and we may have several 100 000 of them)
- test fail for every use of Groebner bases except pure std and liftstd
(for example intersection, quotient, resolution).

I am working on integration of the good ideas, which I see a (in the case of liftstd):
- stop to work on polynomials which turn out to be a syzygy as soon as possible
  (do not enter ist in the list of reducesrs, nad partial GB)
  This is already succesful (08dc13b40d19af0e4a91ab3cd7d095b5884920b7) and 
   reduces compution time by 50 % (in the example)
-  do the reduction as "lazy computation" (as lazy as possible) - most of the reductions tend to 
   lead to syzygies (which are not needed for liftstd). This is more difficult - 
   not just "simple book keeping".

zickgraf:
> but have problems with the real patch:
> 
>     * it nearly doubles the size of TObject (nad LObject) in best case, increase the size in the worst case by 16 kB (and we may have several 100 000 of them)
Yes, of course that's not the way the final solution should look like. I tried to attach the data to `strat` instead (see ffb22cef4dda29c96628d79fdda9b0232e73017c), which I assumed would be the canonical place to store the data, but my C skills were not sufficient to get the memory management correct.
> 
>     * test fail for every use of Groebner bases except pure std and liftstd
>       (for example intersection, quotient, resolution).
Yes, that's expected due to all the edge cases that are missing. In fact, I am surprised that the `liftstd` tests succeed :D
> 
> 
> I am working on integration of the good ideas

Thanks! I will report if I see improvements in my computations :-)

> , which I see a (in the case of liftstd):
> 
>     * stop to work on polynomials which turn out to be a syzygy as soon as possible
>       (do not enter ist in the list of reducesrs, nad partial GB)
>       This is already succesful ([08dc13b](https://github.com/Singular/Singular/commit/08dc13b40d19af0e4a91ab3cd7d095b5884920b7)) and
>       reduces compution time by 50 % (in the example)
I have used this optimization during my tests (its the suggestion I made here: https://github.com/Singular/Singular/pull/965#issuecomment-562643255), but I think this only disguises the problem. If I understand this correctly, one could even remove `IDLIFT` when computing the transformation matrix and syzygies lazily, as then they would be kept in a separate data structure anyway.
> 
>     * do the reduction as "lazy computation" (as lazy as possible) - most of the reductions tend to
>       lead to syzygies (which are not needed for liftstd). This is more difficult -
>       not just "simple book keeping".
Maybe we have different things in mind when talking about "book keeping". For me, "book keeping" by definition includes everything I do to make computations lazy. I will avoid the term in the future.

hannes14:
see 5f5ce616bcb84ae7ec4aa40b04675fe7a3e7e84c, and 9f5bec60f28880de306f761757dc3012518791a1

zickgraf:
@hannes14 Thank you very much for turning this prototype into production code!!! With 5f5ce616bcb84ae7ec4aa40b04675fe7a3e7e84c, `liftstd` seems to be indeed much faster 🎉 Unfortunately, with 817917a7da7faac058cd133c5aef74092b4cf6af we are back at the original speed, so I assume something in that commit is breaking the laziness again :/

hannes14:
should be fixed with https://github.com/Singular/Singular/commit/ddebd75a4ce9e60424c1454a22dee815afe211fe

zickgraf:
That commits does the trick, thanks a lot for the quick fix!

Now I have tried to use it with the relative version of `liftstd` and get a segfault with the following example:
```
LIB "matrix.lib";

ring R = 0,(x,y,z),(c,dp);

matrix A[1][1] = x;
matrix B[1][2] = x+y-z,x-y-z;

matrix C[nrows(A)][ncols(A) + ncols(B)] = concat(A,B);

matrix G1 = std(C);

matrix T;
matrix G2 = liftstd(module(A), T, "std", module(B));

print("");
print("reduce(G1,G2):");
print(reduce(G1,G2));
print("reduce(A*T-G2,std(B)):");
print(reduce(A*T-G2,std(B)));
```
I could fix the segfault (https://github.com/zickgraf/Singular/commit/b114da643beb4812d859b81b7fd9e8d6fa41e9e6) but with the fix I get assume violations:
```
// ***dError: assume violation at ../../../../kernel/GBEngine/kspoly.cc:309 condition: PW->GetpLength() == pLength(PW->p != __null ? PW->p : PW->t_p)
// ***dError: assume violation at ../../../../kernel/GBEngine/kspoly.cc:309 condition: PW->GetpLength() == pLength(PW->p != __null ? PW->p : PW->t_p)
// ***dError: assume violation at ../../../../kernel/GBEngine/kspoly.cc:309 condition: PW->GetpLength() == pLength(PW->p != __null ? PW->p : PW->t_p)
```
and wrong results (the two reductions at the end should both yield 0).

hannes14:
Thanks for the example.
Fixed with 30e184b9f651e8205529c6336cd2f7c39e603699.
Assume violations: in ksReducePoly, only the first part of the polynomials occur,
therefore the length was wrong.

zickgraf:
Again, many thanks for the quick fix! Now my applications look very promising :-)

Unfortunately, I'm running into a new segfault due to `p` being `NULL` here: https://github.com/Singular/Singular/blob/5fd24e69d9cdfcd1040ca93af529563f039efd2c/kernel/GBEngine/kLiftstd.cc#L83

Minimal example:
```
LIB "matrix.lib";

ring R = 0,(x,y),(c,dp);

matrix A[2][1] = x, x;
matrix B[2][1] = y, y;

matrix C[nrows(A)][ncols(A) + ncols(B)] = concat(A,B);

matrix G1 = std(C);

matrix T;
matrix G2 = liftstd(module(A), T, "std", module(B));
```

And I have noticed something weird about the runtime: when choosing large random matrices `A` and `B` in the example above, `liftstd` is actually a bit faster than `std`, which is quite unexpected for me:
```
LIB "random.lib";
// make random reproducible
system("random", 1);
system("--ticks-per-sec",1000);

int t;

int rows = 30;
int columns1 = 350;
int columns2 = 350;

ring R = 0,(x,y,z),(c,dp);
matrix A = randommat(rows, columns1, maxideal(1), 10);
matrix B = randommat(rows, columns2, maxideal(1), 10);
A[1,1] = x^2;

matrix C[nrows(A)][ncols(A) + ncols(B)] = concat(A,B);

t = timer;
matrix G1 = std(C);
print("computed std(C) in " + string(timer - t) + " ms");

matrix T;
t = timer;
matrix G2 = liftstd(module(A), T, "std", module(B));
print("computed liftstd in " + string(timer - t) + " ms");
```
produces
```
computed std(C) in 2220 ms
computed liftstd in 2090 ms
```

But at least until now I could not find an example producing wrong results 🤔 Maybe this is also connected to the segfault.

hannes14:
segfault fixed with 10c7e04a2f5f11b6ae3e82d97926e513c1c821b0

zickgraf:
> segfault fixed with 10c7e04

Thanks! And sorry for always bothering you. Usually I try to fix such issues myself, but with these `NULL`-pointer accesses I am never sure if I should simply return `NULL` or if they need some additional handling.

PR #1041, opened 2021-01-11
Use NEW_ARRAY instead of omAlloc0

mahrud:


mahrud:
56092247878e5b0fff2d73469a4b6324aa23e12e

Issue #1010, opened 2020-06-24
Allow using FLINT instead of NTL for more (all) things


fingolfin:
We should make as much use as possible of FLINT, and ideally make it possible and attractive to use Singular without NTL. That means in all places where we can use NTL right now, we ideally should also allow using FLINT. We should document anything where this is not currently possible and think about how to overcome that.

From https://github.com/oscar-system/SingularBuilder/pull/8:

@hannes14 wrote:
> Beside some experimental stuff (we can live without it: HNF) NTL is used
> for Z-matrices: LLL
> for univariate polynomials::
> 
> * absolute factorization (factoring over Q(alpha) and Z/p(alpa))
> * factorization over Q and Z/p (partially alternative exists)
> * gcd stuff: ezgcd, BuildIrred, ...
> * multiplication over Z/p and Z/p(alpa) (if FLINT is present, FLINT is used)
> * division with/without rest over Z , Z/p, Z/p(alpha) (if FLINT is present, FLINT is used)

@wbhart responded:
> > Beside some experimental stuff (we can live without it: HNF) NTL is used
> > for Z-matrices: LLL
> 
> Flint has HNF and LLL. Both should be comparable to NTL.
> 
> > for univariate polynomials::
> > 
> > * absolute factorization (factoring over Q(alpha) and Z/p(alpa))
> 
> Flint does not have absolute factorisation. I was unaware NTL had it either.
> 
> > * factorization over Q and Z/p (partially alternative exists)
> 
> Flint has this.
> 
> > * gcd stuff: ezgcd, BuildIrred, ...
> 
> I don't know what BuildIrred is. We certainly have GCD, both univariate and multivariate, but I was unaware NTL had ezgcd. Our GCD will be uniformly better.
> 
> > * multiplication over Z/p and Z/p(alpa) (if FLINT is present, FLINT is used)
> > * division with/without rest over Z , Z/p, Z/p(alpha) (if FLINT is present, FLINT is used)
> 
> I guess you mean remainder?

CC @fieker @wdecker

hannes14:
currnet status: without NTL,
-  GCD (but not all variants: need to change the heuristic which to choose)
- univariate factorization
missing: multilvariate factorization

fingolfin:
According to @hannes14, we need need `BuildIrred` and `FindRoot` (for zz_pX/poly over Z/p)
```C
void BuildIrred(zz_pX& f, long n);
// Build a monic irreducible poly of degree n.
[...]

PR #1044, opened 2021-01-25
(try to) fix factorzation over Qa with only flint

tthsqe12:
This seems to work and no more code is missing at least on the examples I tried. Furthermore, the crash in
https://github.com/Singular/Singular/issues/1010#issuecomment-659365767
should be fixed now at least in flint mode.

### DIFF
--- a/factory/FLINTconvert.cc
+++ b/factory/FLINTconvert.cc
@@ -112,16 +112,19 @@ void fq_nmod_set_nmod_poly(fq_nmod_t a, const nmod_poly_t b,
 
 #include "FLINTconvert.h"
 
+// TODO!! decide if the input to convertCF2Fmpz is expected to be:
+//  initialized, or
+//  uninitialized
 void convertCF2Fmpz (fmpz_t result, const CanonicalForm& f)
 {
   if (f.isImm())
-    *result=f.intval();
+    *result=f.intval();                     // assumes uninitialized
     //fmpz_set_si (result, f.intval());
   else
   {
     mpz_t gmp_val;
     f.mpzval(gmp_val);
-    fmpz_set_mpz (result, gmp_val);
+    fmpz_set_mpz (result, gmp_val);         // assumes initialized
     mpz_clear (gmp_val);
   }
 }
@@ -131,7 +134,7 @@ void convertFacCF2Fmpz_poly_t (fmpz_poly_t result, const CanonicalForm& f)
   fmpz_poly_init2 (result, degree (f)+1);
   _fmpz_poly_set_length(result, degree(f)+1);
   for (CFIterator i= f; i.hasTerms(); i++)
-    convertCF2Fmpz (fmpz_poly_get_coeff_ptr(result, i.exp()), i.coeff());
+    convertCF2Fmpz (fmpz_poly_get_coeff_ptr(result, i.exp()), i.coeff()); // assumes initialized
 }
 
 CanonicalForm convertFmpz2CF (const fmpz_t coefficient)
@@ -298,7 +301,7 @@ convertFmpq_poly_t2FacCF (const fmpq_poly_t p, const Variable& x)
 void convertFacCF2Fmpz_array (fmpz* result, const CanonicalForm& f)
 {
   for (CFIterator i= f; i.hasTerms(); i++)
-    convertCF2Fmpz (&result[i.exp()], i.coeff());
+    convertCF2Fmpz (&result[i.exp()], i.coeff()); // assumes ?
 }
 
 void convertFacCF2Fmpq_poly_t (fmpq_poly_t result, const CanonicalForm& f)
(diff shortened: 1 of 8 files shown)


hannes14:
works fine with ntl and ntl+flint, thanks.
without ntl, the following is missing (link errors):
- precomputeLeadingCoeff
- henselLiftAndEarly
- nonMonicHenselLift

tthsqe12:
Yes I see. I am working on it. There is unfortunately a bug in flint that I have to work around as well.

factory/FLINTconvert.cc
```diff
@@ -112,16 +112,19 @@ void fq_nmod_set_nmod_poly(fq_nmod_t a, const nmod_poly_t b,
...
 #include "FLINTconvert.h"
 
+// TODO!! decide if the input to convertCF2Fmpz is expected to be:
+//  initialized, or
+//  uninitialized
 void convertCF2Fmpz (fmpz_t result, const CanonicalForm& f)
 {
   if (f.isImm())
-    *result=f.intval();
+    *result=f.intval();                     // assumes uninitialized
```
tthsqe12, line 121: @hannes14 This is actually an important TODO. The rest of the convertCF2flint functions seem to assume that their inputs are uninitialized (which is not nec. a bad thing), but the majority of the use cases of convertCD2fmpz seems to assume that `result` is initialized. This line `*result=f.intval();` has the potential to leak a lot of integers, and the commented out line `fmpz_set_si (result, f.intval())` would be the correct way if `result` is initialized.

hannes14: I will change it to "result is uninitialized" here and change all uses case which assumes the other way.

hannes14:
still missing: 
- henselLiftAndEarly(CanonicalForm&, List<CanonicalForm>&, int*&, bool&, List<CanonicalForm>&, List<CanonicalForm> const&, List<CanonicalForm> const&, List<CanonicalForm> const&, ExtensionInfo const&)
- nonMonicHenselLift(List<CanonicalForm> const&, List<CanonicalForm> const&, List<CanonicalForm>* const&, List<CanonicalForm>&, Array<CanonicalForm>&, int*, int, bool&)

factory/FLINTconvert.cc
tthsqe12: I would advocate having both since there are a bunch of functions already written assuming initialized input.
```
// as with the other convert functions, result comes in uninitialized
void convertCF2Fmpz (fmpz_t result, const CanonicalForm& f)
{
  if (f.isImm())
    *result=f.intval();
  else
  {
    // promote + demote_val avoids copying the data twice
    fmpz_init(result);
    f.mpzval(_fmpz_promote(result));
    _fmpz_demote_val(result);
  }
}

// special version assuming result is initialized
void convertCF2initFmpz (fmpz_t result, const CanonicalForm& f)
{
  if (f.isImm())
    fmpz_set_si (result, f.intval());
  else
  {
    f.mpzval(_fmpz_promote(result));
    _fmpz_demote_val(result);
  }
}
```

and then the use is

```
void convertFacCF2Fmpz_poly_t (fmpz_poly_t result, const CanonicalForm& f)
{
  fmpz_poly_init2 (result, degree (f)+1);
  _fmpz_poly_set_length(result, degree(f)+1);
  for (CFIterator i= f; i.hasTerms(); i++)
    convertCF2initFmpz (fmpz_poly_get_coeff_ptr(result, i.exp()), i.coeff());
}
```

tthsqe12: I think there should be two functions since a lot of code is currently written
using initialized input:
```
// as with the other convert functions, result comes in uninitialized
void convertCF2Fmpz (fmpz_t result, const CanonicalForm& f)
{
  if (f.isImm())
    *result=f.intval();
  else
  {
    mpz_t gmp_val;
    f.mpzval(gmp_val);

    // promote + demote_val avoids copying the data twice
    fmpz_init(result);
    mpz_swap(gmp_val, _fmpz_promote(result));
    _fmpz_demote_val(result);

    mpz_clear (gmp_val);
  }
}

// special version assuming result is already initialized
void convertCF2initFmpz (fmpz_t result, const CanonicalForm& f)
{
  if (f.isImm())
    fmpz_set_si (result, f.intval());
  else
  {
    mpz_t gmp_val;
    f.mpzval(gmp_val);

    mpz_swap(gmp_val, _fmpz_promote(result));
    _fmpz_demote_val(result);

    mpz_clear (gmp_val);
  }
}
```

and then since fmpz_poly_init2 creates initialized coeffs, the use is

```
void convertFacCF2Fmpz_poly_t (fmpz_poly_t result, const CanonicalForm& f)
{
  fmpz_poly_init2 (result, degree (f)+1);
  _fmpz_poly_set_length(result, degree(f)+1);
  for (CFIterator i= f; i.hasTerms(); i++)
    convertCF2initFmpz (fmpz_poly_get_coeff_ptr(result, i.exp()), i.coeff());
}
```

hannes14:
Merged via command line

PR #1046, opened 2021-02-05
Fixes for liftstd

zickgraf:
@hannes14 I just prepared this PR when I saw your last commit, so sorry for the revert.

`enterOnePairLift` actually assumes `strat->syzComp==1`, see for example:
```
ring R=0,(x,y,z),dp;
matrix A[2][3] = x,x,y,y,z,z;
matrix T;
module B = liftstd(A,T);
```
and compare `std(A)` and `B`: `B` is not a Gröbner basis (as `reduce(std(A),B);` is not zero).

See commit messages for more details.

Edit: maybe one can simply fix `enterOnePairLift` to work with `strat->syzComp>1` but I did not really understand the code.


zickgraf:
Ah, I see that now there is also commit 17bc740f922270721b5e3d90af6f93dfd834b750 which already covers part of this PR. I will open a new PR with only the third commit in this PR.

PR #1047, opened 2021-02-05
liftstd must return a module

zickgraf:
Else information is lost, for example:

ring R = 0,(x,y),lp;
matrix A[2][2] = x,0,0,y;
matrix T;
module S;
print(module(liftstd(module(A),T,S,"std")));

outputs a 1x2 instead of a 2x2 matrix.

I had changed this in 59f0c2b151266af346272ad6c1a729b8294fb6e8 to fix a
test error, but the error actually was that the rank of the result was not
properly set, which is now fixed in bc42722a945eb169cb68ca123321f864dc662fa1


hannes14:
This does not help (see Tst/Short/bug_liftstd4.tst): now liftstd would not return an ideal if input was an ideal.
The real fix is https://github.com/Singular/Singular/blob/spielwiese/Singular/iparith.cc#L7698

zickgraf:
Ah, I see. I did not know the `typeof` command and assumed that a module of rank 1 and an ideal would be more or less the same thing. Thanks a lot for the quick fix! :-)

PR #1048, opened 2021-02-15
Allow to compute relative lifts/syzygies 

zickgraf:
In homalg we often have the situation that we want to compute a relative lift, that is, we want to solve AX=B modulo some relations Y. Equivalently, we want to solve AX+LY = B, where we do not care about the value of Y but only its existence. (This is similar to the situation of `modulo(h1,h2)`: there one wants to compute `S` such that `h1*S` is zero modulo `h2`, that is, `h1*S + h2*Y = 0` for some `Y`.)

Currently, during our algorithm we compute `liftstd( (A|L), T )`, (where `(A|L)` is the union of columns of `A` and `L`), and throw away the last `ncols(L)` lines of `T` afterwards. To speed this situation up, this pull request introduces an optional integer argument to `liftstd`, so we can write `liftstd( (A|L), T, ncols(A) )`. Then in `idPrepare`, instead of putting an identity matrix below `(A|L)` we only put the first `ncols(A)` rows of an identity matrix below `(A|L)`. Depending on the size of `L` this gives a tremendous speedup.

I think this is very similar to what `modulo` does and we can write `modulo(h1,h2)` as `liftstd( (h1|h2), T, S, ncols(h1) )` (+ eliminating some zero lines).

Details regarding the commits:

The first two commits are preparatory commits: The first commit aligns the 4 argument case of `liftstd` with the 2 or 3 arguments case with regard to Letterplace (the 4 argument case was not handled in 82ca97cf422ca0d749d8aaf0e4046e362eb99bdc). The second commit makes the syzygy argument completely optional, in the sense that one can give the string argument without giving the syzygy argument (as we later also want to able to give the integer argument without giving the syzygy argument).

The third commit actually implements the integer argument: I have renamed `jjLIFTSTD_4` to `jjLIFTSTD_M` since it can now be called with 4 or 5 arguments and has to handle all the different cases. `idLiftStd` and `idPrepare` get an optional integer argument which limits the creation of the identity matrix in `idPrepare`. Finally, `T` might now have zero columns, so at the end of `idLiftStd` we actually have to keep track of all the columns not yielding syzygies, even if their lower part is zero.

I have also added a simple test covering all the possible combination of arguments.

References: https://github.com/homalg-project/homalg_project/pull/393 shows how we plan to use this in homalg.


hannes14:
S=liftstd( (A|L), T ,ncols(A)): is identical to T=modulo(A,L): only difference: return of S
S=liftstd(A,T,0) is equivalent to S=std(A)

zickgraf:
> S=liftstd( (A|L), T ,ncols(A)): is identical to T=modulo(A,L): only difference: return of S

Sorry, I don't understand why this should be the case. Consider the following example:
```
> ring R = 0,x,lp;
> matrix A[1][1] = x;
> matrix L[1][1] = 0;
> matrix AL[1][2] = x,0;
> matrix T;
> print(modulo(A,L));
0
> print(liftstd(AL,T,ncols(A)));
x
> print(T);
1,
0 
```
Here `T` is not equal to `modulo(A,L)` (even when only considering the first `ncols(A)` rows of `T`).

If I am not mistaken, then
* we can get `modulo(h1,h2)` from the `S` in `liftstd( (h1|h2), T, S, ncols(h1) )`,
* but we cannot get the `T` in `liftstd( (h1|h2), T, S, ncols(h1) )` from `modulo(h1,h2)`.

For computing relative lifts, we are interested in the `T`.

> S=liftstd(A,T,0) is equivalent to S=std(A)

I  agree! That's a nice special case.

hannes14:
mostly merged manually.

zickgraf:
If I see this correctly, you have changed the integer argument into an ideal/module argument in 17b0ccfd5d571f60393bcf29e604b2e379449352. Now `jjLIFTSTD_SYZ` and `jjLIFTSTD_LIMIT` have the same signature (`MODUL_CMD,  MATRIX_CMD, MODUL_CMD`) and cannot be distinguished anymore, or am I missing something here?

zickgraf:
I have checked out the new implementation and (almost) everything seems to work! I have found two minor issues:
1. The 5 argument version was missing a line, see #1062.
2. `modulo` with three and four arguments does not work as I expect (see below). Since I plan to use `liftstd` and since you still seem to work on the syzygies (judging from the latest commits on `spielwiese`), I have not looked into this. But if you like I can of course debug this further!

Thanks for getting all of this merged! :-)

The issue with modulo:
```
ring R = 0,(x,y,z),lp;
matrix A[2][3] = x,x,y,y,z,z;
module A1=module(A);
module A2=A1[3];
A1=A1[1..2];

matrix T1;
module S;

matrix G=liftstd(A1,T1,S,"std",A2);
print(T1);
print(S);

reduce(G-A1*T1,A2);

matrix T2;

S=modulo(A1,A2,T2,"std");
print(T2);
print(S);

reduce(G-A1*T2,A2);
```

Note that the last columns of `T1` and `T2` differ, and that the last line does not reduce to zero, which is unexpected for me.

PR #1050, opened 2021-02-22
Algorithms for Hodge ideals (hodge.lib)

gblanco92:
This library provides an algorithm to compute Hodge ideals. In particular, the library can also compute multiplier ideals. See the library documentation for more details.

- https://github.com/Singular/Singular/commit/5b749d23d321b83eb9db5b591d0ff5cdc9e84866 fixes a small defect on dmod.lib: the engine type was not forwarded to the function calling `engine`.
- The main function of the library (`Vfiltration`) uses the `modulo` function for some D-modules computation. I have noticed that using `slimgb` in `modulo`  (https://github.com/Singular/Singular/commit/54424d855c3e9062562383d0aa38cd53894787c6) makes it significantly faster (I can provide some benchmark for reference).  https://github.com/Singular/Singular/commit/54424d855c3e9062562383d0aa38cd53894787c6 is not meant to be merged directly into the kernel. However, what would be the correct way to get this same performance from the `modulo` function? This point is quite critical for the code to be usable.
- I have not been able to test this changes with the current `spielwiese` branch (https://github.com/Singular/Singular/commit/b1d05db4b7eb9b213112ad9f8d89f45fd8513c79) because the creation of D-modules (`nc_algebra`) crashes. If you prefer that I report this problem in a separate issue, let me know.

### DIFF
--- a/Singular/LIB/dmod.lib
+++ b/Singular/LIB/dmod.lib
@@ -5024,7 +5024,7 @@ EXAMPLE: example checkRoot; shows examples
   }
   printlevel=printlevel+1;
   def save = basering;
-  def @A = SannfsBM(F);
+  def @A = SannfsBM(F,eng);
   setring @A;
   poly F = imap(save,F);
   number a = imap(save,a);
--- a/Singular/LIB/hodge.lib
+++ b/Singular/LIB/hodge.lib
@@ -1,0 +1,400 @@
+////////////////////////////////////////////////////////////////
+version="version hodge.lib 4.2.0.0 Feb_2021 "; // $Id$
+category="Noncommutative";
+info="
+LIBRARY:  hodge.lib  Algorithms for Hodge ideals
+
+AUTHORS:  Guillem Blanco, email: guillem.blanco@kuleuven.be
+
+OVERVIEW:
+A library for computing the Hodge ideals [MP19] of Q-divisors associated to any reduced hypersurface @math{f \\in R}.
+@* The implemented algorithm [Bla21] is based on the characterization of the Hodge ideals in terms of the @math{V}-filtration of Malgrange and Kashiwara on @math{R_f f^s}, see [MP20].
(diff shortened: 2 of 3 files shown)


levandov (commented):
* ad "ix: missing engine type in dmod.lib" : I totally agree, thanks for this fix!
* ad "trying slimgb in modulo" : I disagree with the replacement. Instead, I propose to duplicate the procedure and make "modulo via slimgb" call available from the interface. Note, that there is a library procedure `moduloSlim` from `nctools.lib' which uses slimgb in the modulo computation on the library level. One could use this name for the kernel call
* hodge.lib: I will read it in details in the next days. Note, that we should think that nearly any groebner-like computation, run in the context of D-modues, should be rather hard. Therefore the usage of dbprint-messages on the progress  is highly recommended.

gblanco92:
> * ad "trying slimgb in modulo" : I disagree with the replacement. Instead, I propose to duplicate the procedure and make "modulo via slimgb" call available from the interface. Note, that there is a library procedure `moduloSlim` from `nctools.lib' which uses slimgb in the modulo computation on the library level. One could use this name for the kernel call

Of course. This replacement was not intended to be included in the kernel as is. It was just to show the performance improvement obtained by switching to `slimgb`. I remember trying the moduloSlim function from `nctools.lib` but the performance was not close to what you obtain using https://github.com/Singular/Singular/commit/54424d855c3e9062562383d0aa38cd53894787c6

> * hodge.lib: I will read it in details in the next days. Note, that we should think that nearly any groebner-like computation, run in the context of D-modues, should be rather hard. Therefore the usage of dbprint-messages on the progress  is highly recommended.

Thanks! I included several dbprint messages in the code. If you think that more are needed just let me know.

hannes14:
First and third commit merged.
I work currently on modulo (with slimgb and more) - this will be solved differently

PR #1063, opened 2021-03-12
CI sage: reduce jobs

mkoeppe:


mkoeppe:
It would probably be good to go to https://github.com/Singular/Singular/actions and cancel some of the workflows that have queued up

PR #1064, opened 2021-03-19
prefer cddlib/ over cdd/

kliem:
An install from source of cdd will put the headers in `cddlib`. So this location folder should be the default place to look for.

Some systems put cdd in `usr/bin/cdd` (e.g. ubuntu bionic). If I have installed an updated cdd somewhere else, it is difficult to link this other one.

This patch will link the cdd installed from source anywhere, if it is early in my $PATH. I would argue that this is what I want in this case.

The downside is, that the system installed cdd will not be chosen anymore (depending on the system), if I also have a user installed cdd in my $PATH. (This could be fixed by enabling only one of `HAVE_CDDLIB_SETOPER_H` etc. according to the first element in my path with cdd.) However, this seems to be an unlikely scenario.

### DIFF
--- a/Singular/dyn_modules/gfanlib/gfan.h
+++ b/Singular/dyn_modules/gfanlib/gfan.h
@@ -13,18 +13,20 @@ Author: monerjan
 #include "misc/int64vec.h"
 
 #include "gfanlib/config.h"
-#ifdef HAVE_CDD_SETOPER_H
-#include <cdd/setoper.h>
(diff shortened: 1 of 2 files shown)


hannes14:
Having the "same" library twice on a system is asking for trouble.
We do not care about the order, just need to find the include files.

kliem:
Thank you. It's a SageMath thing that sometimes system installs are rejected.

PR #1076, opened 2021-04-03
m4/*.m4: In the search loops, test for the presence of headers using the compiler

mkoeppe:
... using `AC_CHECK_HEADERS` instead of just testing the presence of a file.

I have made these changes for FLINT, GMP, and NTL. 
I have not made any changes to `m4/ccluster-check.m4`.


mkoeppe:
Rebased on current `spielwiese` `HEAD`

PR #1094, opened 2021-05-30
Singular/Makefile.am: drop -static linker flags.

orlitzky:
There are a few hard-coded "-static" flags in the Makefile.am for
Singular that are causing weird problems for shared builds. The
visible symptom on Gentoo was a few insecure rpaths being stripped
from the "Singular" executable, but François Bissey was able to track
down the root cause, namely the "-static" that is causing libtool to
act weird.

In retrospect, this problem was also reported on Stack Overflow at,

  https://stackoverflow.com/questions/17905121

The Gentoo bug is,

  https://bugs.gentoo.org/712004

No problems have been experienced in Gentoo after making this change.


orlitzky:
@kiwifb is the one who figured this out

PR #1095, opened 2021-05-30
gfanlib/Makefile.am: use LIBADD for additional libraries.

orlitzky:
The `$GMP_LIBS` and `$CDDGMPLDFLAGS` variables were being added to
`libgfan_la_LDFLAGS`, which isn't quite correct. Since both contain
library (-l) flags, they belong in `_LIBADD` rather than `_LDFLAGS`:

  https://www.gnu.org/software/automake/manual/html_node/Libtool-Flags.html

This commit moves them to a new `libgfan_la_LIBADD` variable.


orlitzky:
This is a rebase of a patch that we have in Gentoo, originally from @kiwifb (?). When certain linker flags like `--as-needed` are used, the order that the flags are passed can become important, and thus does the difference between `_LDFLAGS` and `_LIBADD`.

kiwifb:
Oh, yes I did something like that because libgfan was underlinked and there was no reason for it to be. It is not like there are alternative implementations of cddlib around.

PR #1138, opened 2022-05-20
Make the location of html files configurable

kiwifb:
Since around singular 4.0.x I have been using this patch in Gentoo in some form. The rationale is that linux distribution typically have policy for where the html documentation should be installed. Of course, simply moving the html files in the wanted location will break singular's ability to consult it because the change has to be mirrored in `feResource.cc`. So, this branch create the necessary `HTML_DIR` variable in the appropriate `configure.ac` file and make use of it in `feResource.cc`.

When no `--htmldir` option is passed to configure the previously expected singular default is used. So, in this branch, the behavior is only changed when `--htmldir` is not empty.

Some of the other values in `feResource.cc` may gain to be made more configurable in that way.

### DIFF
--- a/m4/options.m4
+++ b/m4/options.m4
@@ -172,9 +172,15 @@ AC_DEFUN([SING_USE_RESOURCES],
   AC_SUBST(RESOURCES_INCLUDES)
   AC_SUBST(RESOURCES_LIBS)
 
+# If htmldir is the default value, make sure it is passed to resources' configure so that a correct default value is used.
+# if htmldir is set when calling configure, its value is passed by AC_CONFIG_SUBDIRS to resources' configure without manual intervention.
+  if test "x$htmldir" = "x\${docdir}"; then 
+    DEFAULT_HTMLDIR="--htmldir=$datarootdir/doc/singular"
+  fi
+
   ENABLE_ARG="--with-Singular RESOURCES_LIBS='$RESOURCES_LIBS' RESOURCES_INCLUDES='$RESOURCES_INCLUDES'"
 
-  ac_configure_args="$ac_configure_args $ENABLE_ARG"
+  ac_configure_args="$ac_configure_args $DEFAULT_HTMLDIR $ENABLE_ARG"
 
   PKG_REQUIRE="$PKG_REQUIRE singular_resources"
   AC_SUBST(PKG_REQUIRE)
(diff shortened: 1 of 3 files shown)


hannes14:
Nice idea.
But in contrast to the explanation the PR sets HTML_DIR unconditionally (no matter if --htmldir is given or not)
and disables Singulars ability to dynamically determine the location of html files.
(The idea behind this is the possibility to provide binaries without knowing the place of installation).

kiwifb:
Yes, HTML_DIR is set, that is true, and my explanation was based on the idea that the (absolute) location would be the same as before in the default case. I did not take into consideration that scenario of relocation. That means I should replace the current HTML_DIR by a relative path inside `feResource.cc` for that scenario to work. I will give it some thoughts.

kiwifb:
Coming back to this. Would a syntax like
```
{"HtmlDir",   'h',    feResDir,   "SINGULAR_HTML_DIR",    "%D/" + DATA_TO_HTML_DIR,  (char *)""},
```
be acceptable in `feResource.cc`?

hannes14:
Yes, this would keep the possibility of configuration at runtime.

kiwifb:
Actually I was asking about `"%D/" + DATA_TO_HTML_DIR` because I thought putting something like `"%D/$data_to_htmldir"` inside `AC_DEFINE_UNQUOTED` in `configure.ac` wouldn't work properly. But it does. However, the form above may be less mysterious, so I think I will go with it.

kiwifb:
Well it looks like I will have to go to the other form since this one doesn't compile. I mysteriously did try the other one but not that one.

hannes14:
This would search for the html files in <prefix>/share/doc/singular_resources/ if configure is called
without --datadir and/or --htmldir. Possible but unusual. Would require more changes to doc/Makefile.*
for installing the html files.

kiwifb:
That may explain what I thought was a mistake earlier today in the original commit. I'll add an acceptable default.

kiwifb:
This is really tough, autoconf was never meant to be (ab)used that way. It really need to be passed down as a parameter to the resource's configure in any cases.

kiwifb:
After a lot of searching, I got something that work. The key is that `htmldir` has to always be passed to `resources`' configure whether it is set explicitly or not. While I used standard autotools options, it could be passed down with a custom option since it is very specific and is not related to installation of `resources` files but passing a specific location from the parent package. But even in that case, the test to get the default value right would remain.

kiwifb:
I didn't realise comments in the middle of a macro would break autoreconf - so much for trying to put meaningful comments next to the code.

kiwifb:
Should I try to sell you the bits that install the html documentation properly in whatever `htmldir` is set up to?

PR #1143, opened 2022-07-05
Corner case fix for locnormal.lib

kwankyu:
`locnormal` command of `locnormal.lib` errors out if the ideal is already normal. The patch returns the unit ideal with unit denominator for such a case.


kwankyu:
Example:

LIB "locnormal.lib";
ring S = 0,(x,y),dp;
ideal i = y^3 + y - x^4;
locNormal(i);

kwankyu:
Thank you!

PR #1145, opened 2022-07-12
Never use omalloc when it is disabled

catap:


catap:
This is bad idea. omalloc is very heavy integrated and it requires to rewrite to `NEW_ARRAY` a lot of code.

PR #1148, opened 2022-07-17
src: resolve strict-aliasing violation in ndbm.cc

jengelh:
gcc-4.6 warned:

```
ndbm.cc: In function 'datum dbm_nextkey(DBM*)':
ndbm.cc:294:36: warning: dereferencing type-punned pointer will break
strict-aliasing rules [-Wstrict-aliasing]
```


jengelh:
This restores commit f5c16a012d81 which was broken by commit df326b26190fd139d6a4ce8bf0f0001810b5739f

PR #1151, opened 2022-09-28
Fix for non-field base ring

kwankyu:
This is to fix a bug in Singular reported in 

https://trac.sagemath.org/ticket/34417

The patch makes `maPoly_EvalMon` work when the coefficient ring is not a field, such as integer mod ring. 

There is no performance degradation by the patch. The patch was extensively tested through Sage.


kwankyu:
Thank you!

PR #1152, opened 2022-09-28
Update pfd.lib

fchapoton:
just some typos


fchapoton:
should I go on and fix more typos, or do you consider that to be not-so-useful ?

hannes14:
Fix typos if you find them, but no need for an extra search for them.

PR #1159, opened 2022-11-14
trying to add a codespell linter for kernel/

fchapoton:
as an experiment ; should be doable


fchapoton:
ok, now the linter is giving a green light

PR #1177, opened 2023-06-27
make Singular build with flint3

mezzarobba:
This is a naive attempt to make Singular build with flint3. It is
essentially untested apart from checking that it builds with both
flint-2.9.0 and flint-3.0.0-alpha1.


mezzarobba:
Unless I've made a mistake somewhere, this passes `make check` as well as the SageMath test suite with both flint 2.9 and flint 3 alpha1, so I'm marking the PR as ready for review.

Note that some of the changed code could be made a bit faster by using `fmp*_init_set_readonly` and friends, but I think that would break compatibility with old flint versions.

hannes14 (approved):
Thanks

PR #1204, opened 2024-01-30
Improvements and Fixes in rootsur.lib

RickRischter:
1) Variable Reference Correction in Loop:
I modified a for loop to correct the variable reference. Previously, the loop iterated over p[i], but it has been changed to monic[i] to ensure we are accessing the correct element. This change is crucial, especially for handling polynomials with leading coefficients significantly less than 1, such as in cases like p = (x^3)/10-x-1. The previous code could lead to inaccuracies in such scenarios.

2) Proper Calculation of nroots in sturm Routine:
The nroots variable was not initialized, potentially leading to incorrect values in calculations. I have kept the same structure but ensured that the calculation now reflects the correct value.

3) Removal of Unused Static Routines:
I removed the static routines search and D, which were not used elsewhere in the code. 

4) Removal of searchnot Routine Function:
Since the searchnot function was only used by the D routine, which has been removed, I also chose to remove searchnot to simplify the code.

5) Cleanup of Old Code Comments:
I deleted some comments of unused code in the sturmha routine. These comments appear to be remnants of older versions, and their removal makes the code cleaner and easier to understand.

I believe these changes contribute to the efficiency, clarity, and maintainability of the code in rootsur.lib. I am available to discuss these modifications further and to make adjustments as needed.

### DIFF
--- a/Singular/LIB/rootsur.lib
+++ b/Singular/LIB/rootsur.lib
@@ -440,7 +440,7 @@ EXAMPLE:   example maxabs; shows an example"
 
   for (i = 1; i <= size(monic); i++)
   {
-    maximum = max(abs(leadcoef(p[i])),maximum);
+    maximum = max(abs(leadcoef(monic[i])),maximum);
   }
 
   return (maximum + 1);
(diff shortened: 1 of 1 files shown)


hannes14:
Thanks

PR #1249, opened 2024-11-11
Fix bug in map code that didn't handle substitutions efficiently

BrentBaccala:
Commit 5f4616 (Jan 21 2016) introduced code to handle substitutions as a special case within the map code, as a performance optimization

Due to a missing return statement, it didn't actually optimize anything.

This PR puts in the return statement and adds a print statement so that option(prot) will indicate when the optimized substitution map is being used.


hannes14:
Thanks

PR #1250, opened 2024-11-14
Move ksTest_TS() call inside #ifdef KDEBUG

BrentBaccala:
ksTest_* functions are for debugging and should only be called when KDEBUG is defined


BrentBaccala:
typo... I'll reopen another one

PR #1255, opened 2024-11-28
Fixes return call in modstd

ederc:
Fixes wrong return call introduced vai commit 0034a76749121a29e11da98e0f2e904e6761ef70.

@hannes14 Do we really need this? This means we cannot run `modstd` one one core.


hannes14:
modStd creates several processes in the case of 1 core. We may need a modular approach (in the case of 1 core) which runs within the main process, but this is noteeasy to fit into the modular framework.

PR #1262, opened 2025-01-28
chg: replacing Groebner errors with Groebner computations

YueRen:
This is no solution, but a workaround to https://github.com/oscar-system/Singular.jl/issues/846.

Currently Singular has five library functions which error if not given a Groebner basis, all other library functions that require a Groebner basis simply compute it on demand.  This pull request changes the five functions to also compute Groebner bases on demand instead of returning an error.

### DIFF
--- a/Singular/LIB/sheafcoh.lib
+++ b/Singular/LIB/sheafcoh.lib
@@ -232,7 +232,7 @@ KEYWORDS: truncated module
 
   if (attrib(M,"isSB")!=1)
   {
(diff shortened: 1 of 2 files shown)


YueRen:
Sorry, I pushed it to the wrong branch in my repo, please ignore the fact that this PR comes from `yr/tropical`.

ederc (approved):

YueRen:
(minor forced update to fix the spacing, replaces 4 tabs with 4 spaces)

PR #1276, opened 2025-05-29
allow ssh to be used for debugging github test runners

BrentBaccala:
This PR adds to the existing github test harness to provide an option, when the action is triggered manually, to wait at the end for an ssh connection into the test runner, that can be used for debugging.

### DIFF
--- a/.github/workflows/runtests.yml
+++ b/.github/workflows/runtests.yml
@@ -6,6 +6,16 @@ on:
   pull_request:
   workflow_dispatch:
     inputs:
+      repo:
+        description: 'Repository to check out (owner/name); defaults to this repo'
+        required: false
+        type: 'string'
+        default: ''
+      ref:
+        description: 'Branch, tag, or SHA to check out; defaults to the dispatch ref'
+        required: false
+        type: 'string'
+        default: ''
       tmate:
         description: 'At the end, create an ssh target on the test runner and wait for ssh'
         required: false
@@ -30,6 +40,9 @@ jobs:
 
     steps:
       - uses: actions/checkout@v5
+        with:
+          repository: ${{ github.event.inputs.repo || github.repository }}
+          ref: ${{ github.event.inputs.ref || github.ref }}
 
       - name: Install dependencies
         run: |
--- a/.github/workflows/runtests.yml
+++ b/.github/workflows/runtests.yml
@@ -1,5 +1,8 @@
 name: Run tests
 
+# For how to use the workflow_dispatch inputs below (including the
+# interactive-ssh debugging mode), see doc/Debugging-CI.md.
+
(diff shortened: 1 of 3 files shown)


dimpase:
Does this reliably work? We might have a use for it in @sagemath

BrentBaccala:
I haven't used it recently, but it worked fine this spring.  It's a bit of a pain to configure.

To get it working, you need to have the changes made on the github repository's default branch (having it on another branch doesn't work).  You can either change the default branch on the repository settings, or merge the patch into the default branch.  Then you can trigger the workflow manually, as documented [here](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow).

This also means you can't test it normally as a PR.  You can change the git default branch on your own copy of the Singular repository, or merge into it, as I just described, and then make your own PR on your own repository, and run the ssh runner there.  I've done that; it should work.

If the maintainers accept it into the main repository, and put it on the default branch, it will be easier to use.

I was also thinking of using it to diagnose the sagemath test suite failures, as I think you are, but am working on a different project.

Could you try it and see if it works for you?

BrentBaccala:
This PR has been rebased onto current `spielwiese` and expanded with several new commits beyond the original three.

**New workflow inputs** (available when dispatching manually from the Actions tab or `gh workflow run`):

- `repo` / `ref` — check out and test a different repository or branch than the one the workflow lives on
- `tmate` — open an ssh-accessible shell on the runner after tests finish (pass or fail), using [mxschmitt/action-tmate](https://github.com/mxschmitt/action-tmate)
- `tmate_timeout_minutes` — cap how long the session waits for a connection (default 10 min)
- `tmate_limit_access_to_actor` — require the connecting user's ssh key to match the dispatching user's GitHub profile (default true)

Push and pull_request triggers are unchanged — the tmate step is skipped and the checkout behaves as before.

Documentation added in `doc/Debugging-CI.md` covering usage, security considerations, and the tmate upstream deprecation timeline (Homebrew plans to drop the formula on 2026-12-11). A pointer was added from `doc/How-To-Contribute.md`.

**Tested on the fork:**
- Push events correctly skip the tmate step
- Manual dispatch with `tmate=true` reaches the tmate step on all 6 matrix jobs
- `limit-access-to-actor=true`: matching ssh key accepted, non-matching key rejected with "Permission denied (publickey)"

---
*This comment was researched and written by an AI assistant (Claude) on behalf of Brent Baccala (cosine@freesoft.org).*

jankoboehm:
Thanks, we will have a look again, but please don't be disappointed if we currently go for something more standard.

BrentBaccala:
I didn't realize it had been reviewed at all.

PR #1295, opened 2025-09-25
bug fix in posInL110 and posInL110Ring

BrentBaccala:
The ordering goes degree, length, leading monomial.  The leading monomial should be checked only when the lengths are equal, not less than or equal.  The old code used less than or equal and wasn't correct.

This PR is incomplete, as there are failing tests that I haven't completely investigated yet.

The following tests fail only due to changes in the output of `option(prot)`:

- Short/facstd
- Long/homtest
- Old/err3

The following tests exhibit changes in their actual return values and need further review:

- Short/bug_tr707
- Long/eliminate_8


BrentBaccala:
In both `Short/bug_tr707` and `Long/eliminate_8`, the change was a single different polynomial in a basis that passes all other tests.

The PR now includes updated versions of all the changed tests.

PR #1303, opened 2025-11-14
Change syReorderShiftedComponents to static

cxzhong:
fix a compile warning
```
In file included from /usr/include/string.h:548,
from ../../libpolys/misc/intvec.h:9,
from syz1.cc:13:
In function 'void* memcpy(void*, const void*, size_t)',
inlined from 'long int syReorderShiftedComponents(long int*, int)' at syz1.cc:388:9:
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:29:33: warning: 'void* __builtin_memcpy(void*, const void*, long unsigned int)' specified bound between 18446744056529682432 and 18446744073709551608 exceeds maximum object size 9223372036854775807 [-Wstringop-overflow=]
29 | return __builtin___memcpy_chk (__dest, __src, __len,
| ~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
30 | __glibc_objsize0 (__dest));
| ~~~~~~~~~~~~~~~~~~~~~~~~~~
```


cxzhong:
@hannes14 Help me review that. This fixes that compiler regards n could be negative number. it will cause an overflow error.

kernel/GBEngine/syz1.cc
```diff
@@ -333,6 +333,11 @@ SRes syInitRes(ideal arg,int * length, intvec * Tl, intvec * cw)
 // rearrange shifted components
 long syReorderShiftedComponents(long * sc, int n)
 {
+  if (n <= 0)
```
hannes14: syReorderShiftedComponents is only called by syOrder, and here n is always >=1, i.e. N<=0 can not happen

cxzhong: I know that, it just need to tell compiler do not consider the case n<0. So it will not raise the warning

hannes14: Which compiler? Both gcc14/15 and clang with Wall do not complain about syz1.cc.
If really needed, change the type of to unsigned. (and remove syReorderShiftedComponents from syz1.h
anf make it a static routine)

cxzhong:
And in the lastest develop branch
```
In file included from /usr/include/string.h:548,
                 from ../../libpolys/misc/intvec.h:9,
                 from syz1.cc:13:
In function 'void* memcpy(void*, const void*, size_t)',
    inlined from 'long int syReorderShiftedComponents(long int*, int)' at syz1.cc:388:9:
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:29:33: warning: 'void* __builtin_memcpy(void*, const void*, long unsigned int)' specified bound between 18446744056529682432 and 18446744073709551608 exceeds maximum object size 9223372036854775807 [-Wstringop-overflow=]
   29 |   return __builtin___memcpy_chk (__dest, __src, __len,
      |          ~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
   30 |                                  __glibc_objsize0 (__dest));
      |                                  ~~~~~~~~~~~~~~~~~~~~~~~~~~
 ```
 In github CI https://github.com/Singular/Singular/actions/runs/19534266272/job/55924429863

hannes14 (approved):

PR #1306, opened 2025-11-14
Fix typo in lint.yml for Ubuntu runner version

cxzhong:
I am very sorry for that. It is a typo. @hannes14


cxzhong:
Now it is OK for lint. Thank you very much

PR #1309, opened 2025-11-23
cstring include for strlen

dimpase:
building with clang 21 (on Linux) and without omalloc errors out without this change.


dimpase:
@fingolfin - this is a trivial fix. Maybe you could have a look.

hannes14:
I will have a look

hannes14:
seems specific to clang 21 but does not harm other comilers

PR #1315, opened 2026-01-14
Fixes for FGLM using product orderings

ederc:
This is a possible fix due to @jankoboehm for https://github.com/oscar-system/Oscar.jl/issues/5621.

ping: @wdecker


jankoboehm:
@ederc what is the status here? The CI has failures, but that all looks like unrelated CI configuration issues?

Issue #1320, opened 2026-02-13
build system: if `./configure --with-flintlib=PATH` is given and FLINT can't be detected, abort


fingolfin:
Right now, it just continues with FLINT. But if the user explicitly requested it, this is not acceptable.

Motivated by the recent mess on Yggdrasil were Singular failed to find FLINT and we ended up with binaries missing FLINT, without noticing it.

The relevant code is in `m4/flint-check.m4` (the code calling it right now specifies that FLINT should be version >= 2.3 resp. >= 2.4; that, too, should be updated to at least 3.0, possibly more)

PR #1321, opened 2026-02-16
Checking FLINT

ederc:
With these changes `configure` fails if FLINT is requested but not found, see #1320.
This also enforces FLINT version `>=3.0`.


ederc:
ping @fingolfin

jankoboehm:
It seems to me that the script which is called never checks the version number of flint?

fingolfin (approved):
Looks good to me.

Yes it seems to me the FLINT version is never actually checked; but that is a separate issue. I recommend we merge this for now, and the version check could be added in the future, if it is deemed useful. AlternativelY: remove the version argument :-)

PR #1322, opened 2026-02-23
update CI

jankoboehm:
This is work towards updating the CI to 

- run upstream tests of Singular.jl (Oscar.jl in a second step)
- address the failing Sage upstream tests with a less heavy matrix (full matrix still there for manual runs)
- include the `New` tests

### DIFF
--- a/.github/workflows/runtests.yml
+++ b/.github/workflows/runtests.yml
@@ -111,14 +111,14 @@ jobs:
       - name: Install build tools
         run: |
           sudo DEBIAN_FRONTEND=noninteractive apt-get update
+          # NOTE: GMP_jll / MPFR_jll often use the copies bundled with Julia, which do not ship
+          # development headers. We install headers + pkg-config metadata from Ubuntu so that
+          # ./configure can find GMP/MPFR and therefore also FLINT.
           sudo DEBIAN_FRONTEND=noninteractive apt-get install -y \
             autoconf automake libtool pkg-config \
             build-essential \
             curl ca-certificates \
             cmake \
-            # GMP_jll / MPFR_jll usually use the copies bundled with Julia,
-            # which do not ship development headers. We need headers and
-            # pkg-config files to build Singular against FLINT_jll.
             libgmp-dev libmpfr-dev
 
       - name: Set up Julia
--- a/.github/workflows/runtests.yml
+++ b/.github/workflows/runtests.yml
@@ -1,5 +1,6 @@
 name: Run tests
 
+# Trigger the workflow on push or pull request
 on: [push, pull_request]
 
 env:
@@ -21,12 +22,15 @@ jobs:
         run: |
           if [ "$RUNNER_OS" == "Linux" ]; then
             sudo apt-get update
+            # sharutils is for uudecode
             sudo apt-get install -y \
               sharutils libgmp-dev libreadline-dev libmpfr-dev libntl-dev libcdd-dev \
               4ti2 normaliz libopenblas-dev libflint-dev \
               cmake
           elif [ "$RUNNER_OS" == "macOS" ]; then
             brew install autoconf automake libtool gmp readline mpfr ntl cddlib flint
+            # TODO: 4ti2?
+            # TODO: normaliz?
           else
             echo "$RUNNER_OS not supported"
             exit 1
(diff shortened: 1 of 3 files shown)


jankoboehm:
Among the new tests added here (test suit `New`), the following are broken (likely due to the recent regression) and currently deactivated:
Tst/New/dim_integers.tst
Tst/New/qring_ZmtoZmn.tst
Tst/New/qring_ZmtoZmn_errorCoprime.tst
Tst/New/qring_ZtoZm.tst
Tst/New/qring_ntom.tst
Tst/New/qring_with_constant.tst
Tst/New/ringlocal.tst

jankoboehm:
Ping @fingolfin 

Oscar test will still be added.

.github/workflows/runtests.yml
```diff
@@ -16,24 +17,46 @@ jobs:
...
-                    # TODO: normaliz?
-               else
-                    echo "$RUNNER_OS not supported"
-                    exit 1
-               fi
+
+      - name: Install dependencies
+        run: |
+          if [ "$RUNNER_OS" == "Linux" ]; then
+            sudo DEBIAN_FRONTEND=noninteractive apt-get update
```
fingolfin: Why is the `DEBIAN_FRONTEND=noninteractive` suddenly necessary? We don't use it anywhere else

.github/workflows/runtests.yml
```diff
@@ -44,7 +67,272 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+
+      - name: Install build tools
+        run: |
+          sudo DEBIAN_FRONTEND=noninteractive apt-get update
+          # NOTE: GMP_jll / MPFR_jll often use the copies bundled with Julia, which do not ship
+          # development headers. We install headers + pkg-config metadata from Ubuntu so that
+          # ./configure can find GMP/MPFR and therefore also FLINT.
```
fingolfin: Note a good idea.

Better would be to adapt the approach we are taking in https://github.com/oscar-system/Singular.jl/pull/924 and use the `JLLPrefixes` Julia package...

In fact: Perhaps we can wait till that PR is merged, and then use the scripts added there to run the tests. That would simplify code here and have the added benefit of testing those scripts regularly.

CC @lgoettgens @ederc

.github/workflows/runtests.yml
```diff
@@ -16,24 +17,46 @@ jobs:
...
+      #
+      # In addition, to make the upstream test environment match what users of
+      # Oscar/Singular.jl see outside CI, we build a dedicated Singular prefix
+      # against the same Julia JLL dependency stack (GMP_jll, MPFR_jll,
+      # FLINT_jll, cddlib_jll) and using the same configure choices as
+      # Singular_jll (notably: --without-ntl).
+      #
+      # See the Singular_jll build recipe in Yggdrasil for the exact options.
+      # ----------------------------------------------------------------------
+
```
fingolfin: I find this comment, here, surprising.

I support that the tests should "show up as dedicated CI checks" -- for that reason, they should be in a separate .yml file. Perhaps `julia.yml` or so, or even `singular_jl.yml` and later `oscar.yml`.

.github/workflows/runtests.yml
jankoboehm: OK, let's use that script. I suppose that will be merged soon? We should not wait too long to get a working CI here. I will perhaps in the meanwhile adjust this as you suggested, and then we can refer to the script once ready.

lgoettgens: https://github.com/oscar-system/Singular.jl/pull/924 is now merged. I will add a PR for the downstream tests shortly.

lgoettgens: See https://github.com/Singular/Singular/pull/1325

fingolfin:
Reducing the number of sage tests from well over 120 to just 5-6 or so is a very good change. Maybe you can merge that part of this PR right away?

jankoboehm:
OK I will update and merge the Singular and Sage CI changes.

jankoboehm:
@fingolfin is this now ok for you? Then we will merge it.

fingolfin (approved):
Looks good to me

PR #1324, opened 2026-02-25
fixes NULL check in gcd computation for Zn

ederc:
See discussion on https://github.com/oscar-system/Singular.jl/pull/923 .


jankoboehm (approved):
Think we can merge this, also without proper CI running.

PR #1328, opened 2026-02-28
Cleanup gap parentheses

cxzhong:
Cleanup the unnecessary  parentheses. fix the warnings from clang


jankoboehm:
Thank you! Looks good to me. If you like, for clarity, it might now even be better to move the commented out parts out of the conditions, adding instead a line `// Historically: && nCoeff_is_Z(src)`.

cxzhong:
> Thank you! Looks good to me. If you like, for clarity, it might now even be better to move the commented out parts out of the conditions, adding instead a line `// Historically: && nCoeff_is_Z(src)`.

I have fixed this

jankoboehm (approved):

PR #1330, opened 2026-03-03
Add GitHub Actions workflow for releasing tarballs

### LINKED ISSUE #1023: Automate the Singular release process more
Currently making a release of Singular involves a lot of manual work, mostly by @hannes14 . This is IMHO problematic for various reasons:
- Hans' skills are very valuable, and they are IMHO wasted on such a mundane task
- having a high overhead for releases leads to fewer and bigger release; it's difficult to "just make a quick bug fix release"
- mistakes creep in whenever humans have to precisely perform tedious steps (e.g. forgotten tags as in issue #998, versions not properly set in a release tarball as in #987, ...)
- if Hans is not available, we basically cannot make a release (i.e. Singular releases have a "bus factor 1", and we should increase that)

Thus, I believe we should work on automating more of the release, as much as possible. Every step that a well-written script with some good sanity checks performs is one bit less work, and one thing less that can go wrong (again: assuming the script is well-written and handles errors, so some effort is required).

IMHO best is if the full release process is regularly tested as part of the CI; this then can also be used to make "nightly" (or "weekly" or whatever) developer builds available.

Some time ago, Hans sent me two checklists he is using for releases, which I've attached here now. I think there is indeed a lot of potential for more and/or better automation

[MakeSingularRelease.txt](https://github.com/Singular/Singular/files/5007071/MakeSingularRelease.txt)
[wwwsingular.txt](https://github.com/Singular/Singular/files/5007072/wwwsingular.txt)

cxzhong:
I have added a workflow to build a tarball with compiled doc and configured file. So it is easier to create a release package now.

Close #1023 

CC: @ederc @jankoboehm

### DIFF
--- a/.github/workflows/release-tarball.yml
+++ b/.github/workflows/release-tarball.yml
@@ -14,6 +14,8 @@ name: Release Tarball
 on:
   release:
     types: [published]
+  schedule:
+    - cron: '17 9 * * *'
   workflow_dispatch:
     inputs:
       mode:
--- a/.github/workflows/release-tarball.yml
+++ b/.github/workflows/release-tarball.yml
@@ -7,7 +7,7 @@ name: Release Tarball
 ## at https://www.singular.uni-kl.de/ftp/pub/Math/Singular/src/ and used
 ## by SageMath.
 ##
-## Trigger: creating (or editing) a GitHub Release whose tag looks like
+## Trigger: publishing a GitHub Release whose tag looks like
 ##   v4.4.1p5   – patch release  → tarball singular-4.4.1p5.tar.gz
 ##   v4.4.1     – major release  → tarball singular-4.4.1.tar.gz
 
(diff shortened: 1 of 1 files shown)


cxzhong:
run result: https://github.com/cxzhong/Singular/actions/runs/22613451855

jankoboehm:
Thanks! This is very close to the `make_tar.sh` recipe and should solve the “official tarball vs GitHub tarball” issue. It would be great if you could address the following:

1) The header comment says “creating (or editing) a release”, but the trigger is only release: published. Please either adjust the wording or add the respective event types.

2) For attaching assets to a GitHub release, please request explicit token permissions, so this works also under more restrictive defaults. Like in the job that uploads the asset:

```yaml
jobs:
  release-tarball:
    permissions:
      contents: write
```

3) In the tarball step, please enable “fail on pipeline errors” and add simple existence checks before copying/packing the artefacts, so that failures are immediate and easy to diagnose.

cxzhong:
run result https://github.com/cxzhong/Singular/actions/runs/22625882592

cxzhong:
@jankoboehm Can you review this again? It only runs by manunal or release a new version

jankoboehm:
Thanks a lot, we are looking into it, please hang on.

jankoboehm (approved):

PR #1331, opened 2026-03-03
Enable more Singular.jl downstream tests

lgoettgens:
these should work now that Singular.jl 0.28.8 is released.


jankoboehm (approved):
Thank you!

PR #1338, opened 2026-03-18
Fix in hodge.lib

gblanco92:
Fix a bug in hodge.lib: for certain examples, the output could be returned shifted.


Singular/LIB/hodge.lib
```diff
@@ -106,46 +106,33 @@ EXAMPLE:   example Vfiltration; shows an example
...
-    } // Sort shifted eigen.
-  } list S = sort(eigen); eigen = S[1]; mul = mul[S[2]];
-
-  // Remove duplicates & keep max. multiplicity.
-  for (i = 1; i <= size(eigen) - 1; i++) {
-    j = 1; while (eigen[i] == eigen[i+j]) { mul[i] = max(mul[i], mul[i+j]); j++; }
-    while (j != 1) { eigen = delete(eigen, i+j-1); mul = delete(mul, i+j-1); j--; }
+  for (i = 1; i <= size(BS[2]); i++) {
+    if (BS[1][i] < 0) {
+      eigen = eigen+list(number(-BS[1][i]-j));
```
jankoboehm: What is `j` here?

Singular/LIB/hodge.lib
```diff
@@ -106,46 +106,33 @@ EXAMPLE:   example Vfiltration; shows an example
...
+  // Compute the kernels of the eigenvalues in 'eigen' in I.
   str = sprintf("// -3-1- Computing the kernels of the eigenvalues of Dt*t... (p = %p)",
     p); dbprint(ppl, str);
   list Ker; ideal K;
   for (i = 1; i <= size(eigen); i++) {
     str = sprintf("// -3-1-%p- Kernel for eigenvalue %p, multiplicity %p... (%p/%p)",
       i, eigen[i], mul[i], i, size(eigen)); dbprint(ppl, str);
-    // *NO* need to add f^(k-m) as Ker[i] contains J = Annf^(s+m) + f^(k-m).
-    Ker[i] = ideal(modulo(s + eigen[i], J));
+    // *NO* need to add f^(k-m) as Ker[i] contains I = Annf^(s-p) + f^(p+1).
```
jankoboehm: Do you mean `f^(k-m)`?

jankoboehm:
Thanks a lot! Looks like a good improvment. I left a few small comments, please have a look.

jankoboehm:
You mentiion that some results were shifted without the fix. Would it be possible that you add a small regession test exposing the wrong behaviour before the fix and passing now?

gblanco92:
> You mentiion that some results were shifted without the fix. Would it be possible that you add a small regession test exposing the wrong behaviour before the fix and passing now?

How do I write a test in Singular?

Singular/LIB/hodge.lib
gblanco92: Fixed

Singular/LIB/hodge.lib
gblanco92: Fixed

jankoboehm:
Thanks a lot, looks good!

The tests work like this: 

1) You generate a file similar to this:
     https://github.com/Singular/Singular/blob/spielwiese/Tst/Short/Katsura_3_32003_lp_slimgb.tst
     
     named `bug_pr1338.tst` (or what you prefer) with the code which produced the wrong output
     and now is correct.

2) Then generate reference files for the output: If you are in the folder `Tst` do

   `perl ./regress.cmd -s /path/to/your/updated/Singular -g Short/bug_pr1338.tst`
   
    This will generate two files `bug_pr1338.res.gz.uu` and `bug_pr1338.stat` which
    you add to the repo together with the `tst` file. They are used in the tests for reference.
    
3) Finally you add your test file name `bug_pr1338` to one of the list executed in the CI, say

    https://github.com/Singular/Singular/blob/spielwiese/Tst/Short/ok_s.lst
    
    
If you do step one we can also do the rest for you, if you want.

gblanco92:
Added a regression test

Tst/Short/hodge_lib_bugfix1.tst
```diff
@@ -0,0 +1,8 @@
+LIB "tst.lib";
+LIB "hodge.lib";
+tst_init();
+ring MYRINGNAME=0,(x,y,z,w),dp;
+poly MYPOLYNAME=x*w-y*z;
+def Ra=hodgeIdeals(MYPOLYNAME, 3);
```
jankoboehm, line 6: One small thing still: Since `hodgeIdeals` returns a ring, the actual result inside is not printed and hence not fully checked. Could you add something like the following please (and regenerate the reference files):

`setring Ra; hodge;`

gblanco92: Done!

jankoboehm:
Thanks a lot for looking into the test failure. The tests are still failing. As far as I understand, in `hodge.lib`, the printed list is `I_0, I_1, I_2, ...` and for the test hypersurface `x*w-y*z` (ordinary double point in ambient dimension 4), the sequence should be: trivial ideal, trivial ideal, maximal ideal, then its square. In the results file, it seems to me that the comparison is done relative to `(1), (x,y,z,w), ...`, which is the wrong result before the fix. So it looks as if the old Singular version was called by `regress.cmd` to generate the reference file? Did you use the `-s` option to specify the path to the fixed Singular? It will be great if you can have a look (we can also if you prefer).

gblanco92:
Sorry, I was using the wrong Singular in my system to generate the test. Now it should be ok.

jankoboehm (approved):

jankoboehm:
> Sorry, I was using the wrong Singular in my system to generate the test. Now it should be ok.

Thanks a lot!

PR #1353, opened 2026-05-06
Adds weight variants in sheafcoh

ederc:
Adds the functions `dimGradedPart_w` and `dimH_w` which takes as argument corresponding weight vectors. Needed for calling from `Singular.jl` resp. `Oscar`.

Ping: @wdecker


wdecker (dismissed):
thx!

jankoboehm:
Two minor things: Perhaps check w for correct length and change `example dimGradedPart; shows an example` to `example dimGradedPart_w; shows an example`.

jankoboehm (approved):

Issue #1341, opened 2026-04-02
Add WebAssembly as Compiling Target


wangyenshu:
It would be quite interesting to see if it's possible to compile Singular to WebAssembly. That would allow Singular to be used in browser as a static website and also in JavaScript ecosystem.

As a reference, GAP system has a build script for WebAssembly target: https://github.com/gap-system/gap/tree/master/etc/emscripten.

wangyenshu:
I had some experiments with porting Singular to WebAssembly. The build script is here: 
https://github.com/wangyenshu/Singular-wasm/tree/Singular-wasm/emscripten. 
And there is a demo: 
https://wangyenshu.github.io/Singular-wasm/

But many things do not work, like loading lib: `LIB "matrix.lib";` also gfanlib is not included.

jankoboehm:
Thanks a lot for your work! This is quite interesting, we will have a close look and see what could be the remaining issues.

wangyenshu:
Hello. I enabled gfanlib and added normaliz and did some testings. "gfan.lib", "normaliz.lib" and "tropical.lib" are working. I think most library should work. Are there any method to bulk test functionality inside Singular prompt shell?

Also I want to ask that among all depencies of Singular, how to tell which one is shared library and which one is standalone program? Porting shared libraries to WebAssembly is relatively easy and I will do that first.

All testings are performed on https://wangyenshu.github.io/Singular-wasm/. Test logs is here:
[log.txt](https://github.com/user-attachments/files/27564459/log.txt)

jankoboehm:
Thank you very much for your contribution! Starting with it, I have created a browser version of Singular. Your commits are there as well as credit as author on the web frontend, if that is ok for you. This will be in a separate repo, see https://github.com/Singular/Singular-in-browser. Hence I am closing this issue as completed

PR #1360, opened 2026-05-18
add WebAssembly Build Target Support

wangyenshu:
## Description

This pull request adds script and documentations for compiling Singular to WebAssembly. 

A web demo is available on https://wangyenshu.github.io/Singular-wasm/
Compiled Web files: https://github.com/wangyenshu/Singular-wasm/tree/gh-pages

screen

## Testing

Most libraries works, here is the log after running the command `LIB "all.lib";`:
[load_all.log](https://github.com/user-attachments/files/27966048/load_all.log)

## Issue
- A few libraries (like `gitfan`) currently fail to load. A major cause is their hardcoded dependency of `.so` files, which is not possible to load in WebAssembly environment.

- 4ti2 and polymake are not yet available.


jankoboehm:
Thank you very much for your contribution! Starting with it, I have created a browser version of Singular. Your commits are there as well as credit as author on the web frontend, if that is ok for you. This will be in a separate repo, see https://github.com/Singular/Singular-in-browser. Further optimizations will happen there. Accordingly, I am closing this PR and the corresponding issue #1341 as completed (except if there is a fundamental reason that we have to have this in the main repo).

wangyenshu:
That will be great! I am fine with it.

PR #1361, opened 2026-05-20
Adds prune including tracing of reduced components

ederc:
This PR adds the minimal embedding command including the tracing of reduced components but without keeping the map.


ederc:
@wdecker

wdecker (approved):
Thx for doing this!

Issue #1362, opened 2026-05-27
Hilbert driven std computes a Gröbner bases generating an ideal not containing the input ideal


jankoboehm:
Reported on the mailing list:

Dear Singular team,

I would like to report a correctness regression in `groebner` between
Singular 4.1.1 and 4.4.1.  In a quotient ring whose ordering combines a
weighted-matrix `M(...)` block, an `ip(2)` (= `rp(2)` in 4.1.x) block,
and a `dp(7)` block, `groebner` returns an ideal `G` such that `<G>`
does NOT contain all the input generators.  By definition this means
`G` is NOT a Groebner basis of the input ideal.

The same input in Singular 4.1.1 (using `rp(2)` rather than `ip(2)`)
produces a correct Groebner basis.

The four required fields per the guideline at
  https://www.singular.uni-kl.de/index.php/singular-report-bugs.html
follow below.

=========================================================================

Title
-----
`groebner` with mixed `M(...) + ip(2) + dp(7) + C` ordering returns
invalid Groebner basis: three input generators have non-zero normal
form modulo the returned `G`.

Version
-------
Observed:  Singular 4.4.1 (build 44105, 64 bit, May 2026)  INCORRECT
Reference: Singular 4.1.1 (4110, 64 bit, Feb 2018) CORRECT

Component
---------
`groebner` / `std` (= underlying Groebner basis engine), when the ring
ordering combines a weighted matrix block `M(...)` with an `ip(2)` block
and a `dp(7)` block, inside a quotient ring (`qring`).

Code Sample (= standalone, ~40 lines; uses EXPLICIT block sizes to
remove any ambiguity about variable partitioning)
-------------------------------------

  option(noredefine);
  print("Singular version: " + string(system("version")));

  // 12 variables: M covers 3, ip/rp covers 2, dp covers 7.
  if (system("version") < 44000) {
      ring PR = 0, (t, u, v, e(1), e(2), c(1..7)),
                    (M(1,2,6,0,-2,-6,0,-1,0), rp(2), dp(7), C);
  } else {
      ring PR = 0, (t, u, v, e(1), e(2), c(1..7)),
                    (M(1,2,6,0,-2,-6,0,-1,0), ip(2), dp(7), C);
  }
  print("Ringlist (= for cross-version comparison):");
  print(ringlist(basering));

  qring Q = std(ideal(e(1)^2, e(1)*e(2), e(2)^2));

  ideal PRHS =
[...]

PR #1363, opened 2026-05-27
Fix hilbert-driven std regression

jankoboehm:
This should fix the issue #1362 coming from a regression introduced in f565c92bc19a2b5c49a0d6c2b3d48012001c0d99 and 15ad165c6b4f18bddaf2a520e00e0bc5cc6643c2.


Tst/Short/bug_hilbstd_mixed_order.res.gz.uu
fingolfin, line 1: Is there any reason to store these files in uuencoded form? That makes the files harder and makes access less convenient. I can't think of any upside?

(Of course the current setup may *require* this, and so this is not about this PR, but a meta-remark, that perhaps everything should be migrated away from that?

jankoboehm: Currently and since a long time, that is the standard approach. But I agree that it might be good to change that in future. The current setup actually also allows one to provide the res file directly, but when using compression, we require uuencoded. Actually, that was at some point different using just compressed res files. I assume uuencodeding was introduced to make patches and email communication etc easier. Nowadays of course even the compression might be not needed in Github, but locally there is still a benefit for large outputs. So at least uuencode can go. For the moment we perhaps just add a third admissible option for providing the reference files.

wdecker (approved):

PR #1379, opened 2026-08-03
Fix realrad general position normalization

### LINKED ISSUE #1272: realrad runs forever on (x^2-t, y^2+t, z*t-1)
in singular 4.4.0

LIB "realrad.lib";

ring r=0,(x,y,z,t),dp;

ideal i = x^2-t, y^2+t;
ideal j = x^2-t, y^2+t, z*t-1;

realrad(i); 
// quickly and correctly returns the ideal [t, y, x]

realrad(j); // seems to run forever, the correct answer is the ideal [1]

cxzhong:
## Summary

- normalize generators with linear leading monomials before the random general-position change
- recompute each component's Groebner basis in the current variable order
- preserve an invertible random map and restore the saved substitutions after mapping back
- add regression coverage for the original hang, affine relations, and permuted leading variables

## Root cause

`GeneralPos` declared the loop bound `n` without assigning it, so its default value of zero skipped the normalization loop. Simply assigning the variable count is unsafe: the dormant code could remove a coordinate from the random map and lose affine relations. This change rebuilds the maps per component, keeps the random change invertible, and applies the saved inverse substitutions explicitly.

## Impact

The reproducer from #1272 now returns the unit ideal in about 0.45 seconds instead of exceeding a 5-second timeout. Existing `realrad` and `realzero` workloads showed no meaningful performance regression in repeated A/B runs (within about 1.4%).

## Validation

- `Tst/Short/issue1272.tst`
- `Tst/Manual/realrad.tst`
- `Tst/Manual/realzero.tst`
- `Tst/Manual/realpoly.tst`
- `Tst/Short/bug_tr280.tst`
- adversarial matrix: 85/85 public-path cases and 50/50 `GeneralPos` white-box cases
- `git diff --cached --check`

Fixes #1272


jankoboehm (approved):
Thanks! Looks good to me!

PR #1388, opened 2026-08-28
Updates to sresext

jankoboehm:


Tst/Short/sresext.res.gz.uu
Lax202, line 1: What is this file? I don't understand.

jankoboehm: This is the reference file for the tests. The output is compared to what is in this file (in an encoded form).

wdecker (approved):

PR #1394, opened 2026-09-06
Improve the multiplication heuristic

jankoboehm:
Improve the multiplication heuristic internal vs FLINT.


jankoboehm:
Builds on #1393.

PR #1396, opened 2026-09-10
Windows omalloc

jankoboehm:
As an option beyond the compatibility layer (standard for the native windows version), this is a windows native version of omalloc. Currently still a tad slower.


jankoboehm:
Includes #1393.

